PR #220, opened 2015-03-26
localStorage editor mode

waxlamp:
This PR allows Lyra to be booted into editor mode by launching it with a query argument identifying a slot in `localStorage` where the data previously transmitted by `postMessage()` is stored.

This **adds** the `localStorage` functionality without changing the earlier `postMessage()` method.


waxlamp:
@arvind I am no longer using the `postMessage()` approach in Sitar.  If no one else is either, I can add a commit stripping it out.  Otherwise, there is probably little or no harm from leaving it in.

PR #255, opened 2016-02-25
Migrate from Browserify to Webpack for bundle creation

kadamwhite:
The existing build process has several brittle points:
- There is a hard dependency on the file structure inside the `node_modules` folder: this is relied upon to copy out the vendor files
- The vendor file copy script is manually specified and not integrated with the rest of the JS bundling process
- A non-trivial browserify shim configuration is necessary to properly map vendor file `require`s to the appropriate global variables
- The use of browserify shims and global variables can potentially hide unspecified dependencies

Webpack allows us to eliminate the use of shims, and instead instruct our JS build to separate the code into two bundles: a vendor bundle, and our bundle of first-party code. While this causes the build process to take a few seconds longer when run on its own via `webpack -d`, when using `--watch` the `vendor.js` bundle is written once and henceforth ignored, so _re_-builds of the first-party Lyra code execute very quickly. This configuration is simultaneously more powerful and more readable than the Browserify equivalent, and Webpack sets us up to leverage hot module replacement later on.

Webpack also reduces the size of the application _vs_ browserify, and will reduce it even more when run in production mode:

Browserify: 2.5mb of JS

> Browserify build JS network statistics

Webpack: 2.1mb of JS

> image

For LYRA-252


deathbearbrown:
This looks really clean.   We should probably add an .editorconfig with the linter so we don't end up having PRs with files where only spacing changes.

kadamwhite:
Because this switches us away from `React` being a global, there's a possibility of regressions. I've caught the most obvious ones but this could use some good spot-testing.

.gitignore
```diff
@@ -4,6 +4,7 @@ _site
 bower_components
 coverage
 node_modules
+*.log
 output
 temp
-vendor
\ No newline at end of file
+build
```
kadamwhite, line 10: This also accomplishes the `vendor/` -> `build/` switch we discussed yesterday

package.json
```diff
@@ -54,53 +51,20 @@
...
-    "lint:js": "jshint --verbose src/js/",
-    "vendor": "node ./scripts/vendor.js",
     "build": "npm run build:js && npm run build:scss",
-    "build:js": "browserify src/js/index.js -d -o vendor/js/lyra.js",
-    "postbuild:js": "uglifyjs vendor/js/lyra.js -cm > vendor/js/lyra.min.js",
-    "build:scss": "node-sass --include-path src/scss src/scss/main.scss vendor/css/main.css",
-    "postbuild:scss": "cleancss -o vendor/css/main.min.css vendor/css/main.css",
+    "build:js": "webpack -d",
+    "postbuild:js": "uglifyjs build/js/lyra.js -cm > build/js/lyra.min.js",
+    "build:scss": "node-sass --include-path src/scss src/scss/main.scss build/css/main.css",
```
kadamwhite, line 60: Long-term we may benefit from switching from explicitly calling UglifyJS to use Webpack production mode instead, but for now I opted to maintain the division of labor.

src/js/components/index.jsx
```diff
@@ -1,7 +1,8 @@
 var ReactDOM = require('react-dom'),
-    Sidebars = require('./Sidebars.jsx');
+    Sidebars = require('./Sidebars.jsx'),
+    React = require('react');
 
 module.exports = ReactDOM.render(
   <Sidebars />,
```
kadamwhite, line 6: Anywhere we used JSX we must explicitly require React `var React = require('react');`, because the React loader considers using a component via JSX syntax is equivalent to calling React's method to create that element.

deathbearbrown:
+1 to code.  Anything I would have to say is probably getting fixed when we do the linting, so I'll hold off.

Functional testing, I have been messing with everything and it doesn't seem to have broken things

I don't know if tests worked before, `npm run test` isn't working for me right now.  I know you are going to add some testing coverage with istanbul, so that's probably going to be handled in another PR.

PR #260, opened 2016-02-26
WIP: Add Jsdocs

deathbearbrown:
### Description of the problem this pull request fixes:

Adding JS doc and a minimal theme less hideous than the regular js doc theme
- run `npm run docs`
- open simple server
- got to:  http://localhost:8000/docs

There isn't much here. I plan on going in and adding place holders around all the areas we need to add more docs.

I also am trying to think of the best way to handle design consistency with the jekyll site. I might fork the theme over the weekend and then update it to be prettier. 

screen shot 2016-02-26 at 3 37 55 pm


deathbearbrown:
@arvind said he has time tonight to brain dump into the docs. I found their documentation ironically to be awful. This site helped me more: http://www.2ality.com/2011/08/jsdoc-intro.html

Also check out the moebius project:
compiled docs: http://moebiolabs.github.io/moebio_framework/docs/Axis2D.html
Here's an example of documenting a class: https://github.com/moebiolabs/moebio_framework/blob/master/src/dataTypes/geometry/Point.js

deathbearbrown:
Some docs I made for contributing documentation:

## Using JS Doc

Any comments written with a comment block `/** comment block **/` will get picked up by jsdoc.  If you have comments you would like to make that don't show up in the documentation, use `// comment`.

Namespace docs are created by putting `/** @namespace */` before the variable. 
Each key can be documented with a comment block that is broken down like this:

```
/**
 * Brief description of the method
 * @param {number} n number of things
 * @param {string} propName - Some long description of what this does
 * that really is too complex to represent on one line (when it is
 * better to be specific than terse)
 * @returns {type}
 */
```

Type can be string, object, integer, number, boolean, or even a class.
Primitive types are boolean, number, string. The values undefined and null are also considered primitive. These are always lowercase. Object types are Objects including Arrays and Functions and are written uppercase.

Example:

```
 /** @namespace */
    var util = {
        /**
         * Repeat <tt>str</tt> several times.
         * @param {string} str The string to repeat.
         * @param {number} [times=1] How many times to repeat the string.
         * @returns {string}
         */
        repeat: function(str, times) {
            if (times === undefined || times < 1) {
                times = 1;
            }
            return new Array(times+1).join(str);
        }
    };
```

### Defining a Class

Before the function that sets up the class add a code block that contains `@class` or `@constructor`.

Example:

```
/**
 * Represents a Primitive.
 * @constructor
 * @param {string} value description
 * @property {string} name 
 */
function Primitive(value) {
  this._name = value;
  model.primitive(this._id=++id, this);
  return this;
}

/**
 * Returns the primitive name
 * @returns {string}
 */
Primitive.prototype.getName = function(){
  return this._name;
}
```

Setting it up this way makes it compile with all the class methods on the same page.

If you compile and for some reason a method ends up under `global`, you can force it to be under a class by using `@memberof Class` or `@memberof! Class`. The ! is similar to using an !important tag in css. 

If a class extends another class, you can document that with `@extends Class`.

You can chose to document private and public variables using `@private` and `@public`.  With our conventions, private variables will be referenced with an underscore. eg: `this._foo = secrets;`.

.jsdoc.json
```diff
@@ -0,0 +1,25 @@
...
+        "monospaceLinks": true
+    },
+    "opts": {
+        "destination": "./docs/",
+        "encoding": "utf8",
+        "private": true,
+        "recurse": true,
+        "template": "./node_modules/minami"
+    }
+}
```
kadamwhite: editorconfig should be catching files without EOF newline

src/js/model/primitives/Guide.js
```diff
@@ -1,9 +1,18 @@
...
 var GTYPES = {AXIS: 1, LEGEND: 2},
     ORIENT = {x: 'bottom', y: 'left'};
+/**
+* Guide
+* @constructor
+* @param {string} gtype
+* @param {string} type
+* @param {string} scale
+*
+*/
```
kadamwhite: Do we care about having a preceding space?

``` js
/**
 * Something
 */
```

vs

``` js
/**
* something
*/
```

I am not sure whether this is enforceable via eslint

src/js/model/primitives/Guide.js
```diff
@@ -1,9 +1,18 @@
...
     lookup = model.primitive;
 
 var GTYPES = {AXIS: 1, LEGEND: 2},
     ORIENT = {x: 'bottom', y: 'left'};
+/**
+* Guide
+* @constructor
+* @param {string} gtype
+* @param {string} type
+* @param {string} scale
```
kadamwhite: What's our preference for handling spacing on long comments?

``` js
/**
 * @param {string} propName - Some long description of what this does
 * that really is too complex to represent on one line (when it is
 * better to be specific than terse)
 */

/**
 * @param {string} propName - Some long description of what this does
 *                            that really is too complex to represent
 *                            on one line (when it is better to be
 *                            specific than terse)
 */
```

src/js/model/primitives/Guide.js
```diff
@@ -1,9 +1,18 @@
...
 var Primitive = require('./Primitive'),
     model  = require('../'),
     lookup = model.primitive;
 
 var GTYPES = {AXIS: 1, LEGEND: 2},
     ORIENT = {x: 'bottom', y: 'left'};
+/**
+* Guide
+* @constructor
+* @param {string} gtype
```
kadamwhite: is there a nuance between `string` and `String`?

src/js/model/primitives/Guide.js
deathbearbrown: lets do the top one since the sublime package does that by default :)

src/js/model/primitives/Guide.js
deathbearbrown: I prefer the top one

src/js/model/primitives/Guide.js
kadamwhite: Awesome!

Speaking of the sublime plugin, @arvind if you use Sublime, the DocBlockr extension is great: it auto-completes and lets you tab through a snippet. I'm typing `/**` and hitting Enter, then tabbing
dockblockr

src/js/model/primitives/Guide.js
deathbearbrown: I'll add to the docs about this :)

deathbearbrown:
@kadamwhite cleaned it up :)

src/js/transforms/DropZone.js
```diff
@@ -7,9 +7,27 @@ var dl = require('datalib'),
     Transform = vg.Transform,
     sg = require('../model/signals');
 
+
+DropZone.prototype = Object.create(Transform.prototype);
+DropZone.prototype.constructor = DropZone;
```
kadamwhite, line 12: This only works because of hoisting -- that'll probably be caught by the linter so it doesn't need to be addressed here, but as a note for @arvind we should avoid extending constructors before they're defined ("before" meaning as we read linearly through a file).

src/js/transforms/DropZone.js
```diff
@@ -19,10 +37,12 @@ function DropZone(graph) {
...
-var prototype = (DropZone.prototype = Object.create(Transform.prototype));
-prototype.constructor = DropZone;
-
-prototype.transform = function(input) {
+/**
+ * transform dropzone
+ * @param {string} input - todo.
+ * @return {object} output - todo
+ */
+DropZone.prototype.transform = function(input) {
```
kadamwhite, line 45: @arvind this may have been discussed but we prefer `ConstructorFn.prototype.fnName =` to using an intermediate variable for obviousness and to avoid having to manually specify as much information via JSDoc (the more we can let JSDoc's "magic" do its thing, the easier it is to maintain the documentation)

src/js/util.js
```diff
@@ -4,19 +4,38 @@ var dl = require('datalib'),
     NS = 'lyra_',
     vgSchema;
 
-function ns(name) {
-  return name.startsWith(NS) ? name : NS + name;
-}
+function ns (name) {
+  return name.startsWith(NS) ? name : NS+name;
+};
```
kadamwhite, line 9: Think this `;` is superfluous but again, fixing it isn't relevant to this PR

PR #261, opened 2016-03-01
Consider removing .jsx extensions from require() calls

kadamwhite:
By default React eslint rules flag require calls that specify `.jsx`: the rationale from the [`eslint-plugin-react` documentation](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-extension.md) is that,

> `require()` statements should generally not include a file extension as there is a well defined mechanism for resolving a module ID to a specific file.

Our Webpack configuration does handle this resolution so it is accurate that the extension is unneeded.

Removing the .jsx extensions from require calls suppresses 54 errors, but I do not want to make the assumption that changing the code is the way to go: we could also suppress this eslint rule.

There does not seem to be a universal preference about using or omitting the extension, but the majority of coworkers I polled have adopted the extensionless form.

Thoughts on this, @arvind @deathbearbrown?

LYRA-261

### DIFF
--- a/src/js/components/ContentEditable.jsx
+++ b/src/js/components/ContentEditable.jsx
@@ -2,7 +2,7 @@
 var dl = require('datalib'),
     React = require('react'),
     ReactDOM = require('react-dom'),
-    SignalValueMixin = require('./mixins/SignalValue.jsx'),
+    SignalValueMixin = require('./mixins/SignalValue'),
     model = require('../model');
 
 var ContentEditable = React.createClass({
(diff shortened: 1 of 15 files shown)


arvind:
+1 for removing the `jsx` extension. I was never happy that they were present -- an artifact of the fact that I couldn't get the old browserify build system to resolve them w/o the extension.

kadamwhite:
OK, awesome! Thanks for the historical perspective.

PR #263, opened 2016-03-02
[WIP] JSDocs

arvind:
This adds documentation for the majority of the non-JSX source code. There are three pending areas that need to be documented:
- [ ] `model/signals`
- [ ] `model/rules`
- [ ] `model/primitives/marks/manipulators`

For the latter two, I wasn't entirely sure which JSDoc tag I should be using since both of those extend the `Mark` prototype. It's not exactly an `@interface` and using `@memberof` suggests that they're static members, which they're not. 

It might, of course, be a sign that there's no good JSDoc tag to describe that design pattern. So perhaps they need to be refactored and moved into `Mark` proper? The reason they are where they are now, though, is I felt it better encapsulated functionality (i.e., all `rules` logic together, and all `manipulators` logic together rather than spread across different modules).

### DIFF
--- a/src/js/model/primitives/Primitive.js
+++ b/src/js/model/primitives/Primitive.js
@@ -4,18 +4,24 @@ var dl = require('datalib'),
     lookup = model.primitive,
     id = 0;
 
+/**
+ * @classdesc A Lyra Primitive.
+ *
+ * @description Lyra primitives are wrappers around corresponding Vega
+ * definitions. Lyra-specific properties are denoted with an underscore prefix
+ * and are removed during clean exports. Each Lyra primitive is given a unique
+ * identifier, and these identifiers are used to store references to other
+ * primitives.
+ *
+ * @constructor
+ */
 function Primitive() {
   model.primitive(this._id = ++id, this);
   return this;
 }
 
-var prototype = Primitive.prototype;
-
-// Primitive classes are wrappers around the corresponding Vega
-// specification. Clean them up to remove Lyra-specific things.
-// This also converts signalRefs -> values unless resolve === false.
-function clean(spec, resolve) {
-  var k, p, c, res = resolve !== false;
+function _clean(spec, clean) {
+  var k, p, c, cln = clean !== false;
   for (k in spec) {
     p = spec[k];
     c = k.startsWith('_');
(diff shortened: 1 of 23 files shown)


src/js/model/primitives/Primitive.js
```diff
@@ -4,18 +4,24 @@ var dl = require('datalib'),
...
 
-var prototype = Primitive.prototype;
-
-// Primitive classes are wrappers around the corresponding Vega
-// specification. Clean them up to remove Lyra-specific things.
-// This also converts signalRefs -> values unless resolve === false.
-function clean(spec, resolve) {
-  var k, p, c, res = resolve !== false;
+function _clean(spec, clean) {
+  var k, p, c, cln = clean !== false;
```
kadamwhite, line 24: I have a strong bias against one-letter variable names -- variables in a perfect world are self-documenting, but single-letter identifiers always require mental tracking as you read the code. Given that this removed the comment around this function I'd prefer to see the variable names made more readable.

src/js/model/primitives/Primitive.js
```diff
@@ -24,18 +30,48 @@ function clean(spec, resolve) {
...
+ * @return {Object} A Vega specification.
+ */
+Primitive.prototype.export = function(clean) {
+  return _clean(dl.duplicate(this), clean);
+};
+
+/**
+ * Exports the primitive as a complete Vega specification with extra definitions
+ * to power Lyra-specific interaction (e.g., extra manipulator mark definitions).
+ * @return {Object} A Vega specification.
```
kadamwhite, line 62: @deathbearbrown when returning an instance should we be using

``` js
/**
 * @return {Primative}
```

?

deathbearbrown, line 62: Return type

kadamwhite, line 62: I don't understand that response, sorry—could you clarify?

deathbearbrown, line 62: Types are made from primitives.  The common language to use is @return {type}, in this case @return {Object} is correct.

kadamwhite, line 62: OK, that's a departure from how I'd interpreted it—given that we can specify a constructor in our `@param` definitions or [type specifications](http://usejsdoc.org/tags-type.html), I was expecting consistency through `@returns`. Put another way, nothing in the jsdoc docs I can find implies to me that this wouldn't be preferable:

``` js
/**
 * @returns {MyClass}
 */
function returnClass() { return new MyClass(); }

/**
 * @param {MyClass} classInstance
 */
function doSomethingWithInstance( classInstance ) { classInstance.everything(); }
```

arvind, line 62: Thanks for this, I wasn't sure what the convention was either. jsdoc3/jsdoc#991 suggests that this isn't something the JSDoc team is interested in supporting, unfortunately, and the array of suggestions didn't give me confidence that there was a standard which is why I defaulted back to JavaScript types + descriptions.

src/js/model/primitives/Primitive.js
arvind, line 24: Good call on variable names. I thought I had incorporated this comment in the class descriptor + `export` method but looking at it now, that's not entirely true. Must've spaced last night. Will fix presently.

src/js/model/primitives/Primitive.js
kadamwhite, line 62: I believe that's a related but separate issue -- `@return {this}` would be useful for chaining, but that's actually a step beyond what I was advocating for / wondering why it didn't seem to be "supported," per @deathbearbrown. If I see `@returns {Object}` I would expect it to be either an object literal or a `new Object()`. If it returns an instance of a constructor, with all the functionality that being an instance implies, I am used to (from YUIdoc) `@returns {MyClass}`. If JSDoc really provides no facility to specify a custom constructor as its return type I am both disappointed and surprised.

kadamwhite, line 62: Per discussion in standup, capturing this for posterity: This is correct, use the class name as the type when specifying `@return`. Err on the side of specificity.

kadamwhite:
I've created issue #269 to capture the remaining TODOs; I will rebase and merge this

PR #266, opened 2016-03-02
Remove unused package deps & build scripts

kadamwhite:

Many dependencies specified in package.json are not in use within Lyra, and not all of the defined package scripts have the desired (or any) effect.

**Changes proposed**

This patch removes all NPM dependencies that are not obviously being used in the application.

It also removes the infinitely-recursing `npm test` script, and nixes the two `postbuild:` steps: the application is not currently set up to handle minified files, and we will be implementing minification via the webpack bundler rather than calling uglify directly when we do get to that point.

**Steps to functionally test**
1. Remove your `node_modules` directory
2. Reinstall the application
3. Confirm that Lyra builds
4. Load the dev server and confirm Lyra works as expected

LYRA-262, LYRA-263


package.json
```diff
@@ -61,13 +53,10 @@
     "lint": "eslint --ext .js,.jsx src",
     "build": "npm run build:js && npm run build:scss",
     "build:js": "webpack -d",
-    "postbuild:js": "uglifyjs build/js/lyra.js -cm > build/js/lyra.min.js",
     "build:scss": "node-sass --include-path src/scss src/scss/main.scss build/css/main.css",
-    "postbuild:scss": "cleancss -o build/css/main.min.css build/css/main.css",
```
kadamwhite, line 66: I've created #267 to track adding this minification back in; our dev server and build process don't currently account for the minified files so these steps are currently just baggage, hence removing them rather than simply leaving them be.

package.json
```diff
@@ -61,13 +53,10 @@
...
     "watch": "npm run watch:js & npm run watch:scss",
     "watch:js": "webpack -d --watch",
     "watch:scss": "nodemon -e src/scss -x \"npm run build:scss\"",
-    "test": "npm run test",
```
kadamwhite, line 70: Assuming @deathbearbrown will reinstate this in such a way that it actually does something as she gets Karma set up

package.json
deathbearbrown, line 66: we... need this one I think hehe

screen shot 2016-03-02 at 1 33 17 pm

package.json
deathbearbrown, line 70: :+1:  I am on it.   (also this didn't work before)

deathbearbrown:
The css needs to get fixed to merge this.  Just so it's not jarring for anyone doing development on the app to have a css-less site.

package.json
kadamwhite, line 66: AHA you are correct. Gon' re-write the HTML to reference `main.css` instead of the minified file, for now.

kadamwhite:
@deathbearbrown CSS issue fixed

deathbearbrown:
+1 mergin

PR #271, opened 2016-03-03
Support sourcemaps in webpack-dev-server

kadamwhite:
**Description of the problem this pull request fixes**

I neglected to set the webpack dev server to output sourcemaps, which renders it less than useful for development and debugging purposes. Based on the matrix of utility to build performance for each sourcemap method supported by the server, I've opted to implement the "cheap module source map" option, which provides line numbers but not column information within a line.

[Matrix of webpack-dev-server devtool options](https://webpack.github.io/docs/configuration.html#devtool), with selected option called out in bold:

| devtool | build speed | rebuild speed | production supported | quality |
| --- | --- | --- | --- | --- |
| eval | +++ | +++ | no | generated code |
| cheap-eval-source-map | + | ++ | no | transformed code (lines only) |
| cheap-source-map | + | o | yes | transformed code (lines only) |
| cheap-module-eval-source-map | o | ++ | no | original source (lines only) |
| **cheap-module-source-map** | **o** | **-** | **yes** | **original source (lines only)** |
| eval-source-map | -- | + | no | original source |
| source-map | -- | -- | yes | original source |

PR #272, opened 2016-03-03
Test

deathbearbrown:

Adding Karma, Mocha, Chai, Sinon, Enzyme stack for testing and jsx template support.

Right now this just has a hello world script and the other scripts are commented out.  I started fixing the other scripts, but realized that the model is running before everything runs, and I know we are planning to use redux (or something) and change the model soon.  I'd rather wait until that happens to fix the scripts.

To test:
- npm install
- npm test

Also test for regressions:
- npm run build
- npm start


karma.config.js
```diff
@@ -0,0 +1,70 @@
...
+
+    // include some polyfills for babel and phantomjs
+    files: [
+      'node_modules/babel-polyfill/dist/polyfill.js',
+      './node_modules/phantomjs-polyfill/bind-polyfill.js',
+      './test/**/*.js' // specify files to watch for tests
+    ],
+    preprocessors: {
+      // these files we want to be precompiled with webpack
+      // also run tests throug sourcemap for easier debugging
```
kadamwhite: 'through'

karma.config.js
```diff
@@ -0,0 +1,70 @@
...
+      ['./test/**/*.js']: ['webpack', 'sourcemap']
+    },
+    webpack: {
+       devtool: 'inline-source-map',
```
kadamwhite, line 25: After reading up more on devtools for #271 I want to do a speed test b/w this and `cheap-module-inline-source-map`

package.json
```diff
@@ -23,27 +23,44 @@
...
+    "babel-preset-react": "^6.3.13",
+    "babel-register": "^6.3.13",
     "chai": "^3.0.0",
-    "chai-spies": "^0.7.1",
     "d3": "^3.5.6",
     "d3-geo-projection": "^0.2.15",
     "datalib": "^1.4.6",
     "deep-diff": "^0.3.3",
+    "enzyme": "^1.2.0",
+    "es6-promise": "^3.0.2",
```
kadamwhite: This doesn't look like it's used, may be extraneous

package.json
```diff
@@ -23,27 +23,44 @@
...
     "es6-promisify": "^3.0.0",
     "eslint": "^1.10.3",
     "eslint-loader": "^1.2.1",
     "eslint-plugin-react": "^3.16.1",
+    "fs": "0.0.2",
     "istanbul": "^0.3.21",
     "jshint": "^2.8.0",
+    "json-loader": "^0.5.4",
+    "jstransform": "^11.0.3",
```
kadamwhite: This is another one that was removed in #266; it's unclear whether it was added back in to support this build or due to a merge with old code

test/test_helper.js
```diff
@@ -0,0 +1,44 @@
+// no longer being used once Karma is set up
+var expect = require('chai').expect,
+    sinon = require('sinon'),
+    fs = require('fs'),
+    res = './test/resources/';
+
+global.expect = expect;
+global.sinon = sinon;
```
kadamwhite: Do these need to be global?

kadamwhite:
@arvind It looks like on other Vega projects the team switched from `filename.spec.js` to `filename.test.js` -- This goes the other direction, do you have a strong preference either way?

arvind:
`filename.test.js` would be my preference for consistency.

kadamwhite:
@deathbearbrown :+1: from me on regression & testing, this is working -- thanks so much for taking on setting this up, I know these tools took a while to configure! I have some naive questions in the comments above, and Arvind's preference is to maintain `.test.js` as the extension convention, but this should be able to :ship: with those changes.

package.json
```diff
@@ -58,7 +75,8 @@
     "watch": "npm run watch:js & npm run watch:scss",
     "watch:js": "webpack -d --watch",
     "watch:scss": "nodemon -e src/scss -x \"npm run build:scss\"",
-    "test:js": "mocha --timeout 30000 --recursive test/",
+    "test": "node_modules/.bin/karma start karma.config.js",
```
arvind: `npm` automatically puts `node_modules/.bin` in the path for `npm run` scripts. So you should be able to define this as `karma start karma.config.js` directly.

package.json
deathbearbrown: json-loader should be there I think jstransform was a dependency of babel-es2015 which I removed

package.json
deathbearbrown: yeah, that's a dependency of something I removed >.<

deathbearbrown:
updated to file.test.js

karma.config.js
```diff
@@ -0,0 +1,70 @@
...
+    // include some polyfills for babel and phantomjs
+    files: [
+      'node_modules/babel-polyfill/dist/polyfill.js',
+      './node_modules/phantomjs-polyfill/bind-polyfill.js',
+      './test/**/*.js' // specify files to watch for tests
+    ],
+    preprocessors: {
+      // these files we want to be precompiled with webpack
+      // also run tests through sourcemap for easier debugging
+      ['./test/**/*.js']: ['webpack', 'sourcemap']
```
arvind: I get an error thrown on this line: 

```
03 03 2016 14:53:37.940:ERROR [config]: Invalid config file!
  /Users/arvind/Research/lyra/code/karma.config.js:22
      ['./test/**/*.js']: ['webpack', 'sourcemap']
      ^
SyntaxError: Unexpected token [
```

Should this be valid JSON (`'./test/**/*.js'`)?

karma.config.js
```diff
@@ -0,0 +1,70 @@
+var argv = require('yargs').argv;
```
arvind, line 1: `yargs` is also not defined as a `devDependency` in `package.json`.

karma.config.js
deathbearbrown: hrmm... lemme see if I can recreate that.  Is this after running `npm test` ?

arvind:
If I resolve my two inline comments and then `npm run test`, I see 

```
PhantomJS 1.9.8 (Mac OS X 0.0.0) LOG LOG: 'I'm `fs` modules'

  hello world
    ✓ works!

PhantomJS 1.9.8 (Mac OS X 0.0.0): Executed 1 of 1 SUCCESS (0.003 secs / 0 secs)
TOTAL: 1 SUCCESS
```

Except, in the prior output, there is 

```
ERROR in ./src/js/components/index.jsx
Module build failed: SyntaxError: /Users/arvind/Research/lyra/code/src/js/components/index.jsx: Unexpected token (7:2)
   5 |
   6 | module.exports = ReactDOM.render(
>  7 |   <Sidebars />,
     |   ^
   8 |   d3.select('#sidebars').node()
   9 | );
  10 |
    at Parser.pp.raise (/Users/arvind/Research/lyra/code/node_modules/babel-core/node_modules/babylon/index.js:1412:13)
...
```

Is it problematic for the test runner to return a successful result if there was an intermediate build failure even if it's not testing that file specifically?

karma.config.js
deathbearbrown, line 1: it's a dependency of webpack.

arvind, line 1: Hmm, I'm on node `0.12.10` and npm `2.14.9` and on clean `npm install` of this branch a subsequent `npm run test` throws this error (assuming the above config error is fixed). 

```
03 03 2016 15:19:39.578:ERROR [config]: Invalid config file!
  Error: Cannot find module 'yargs'
```

deathbearbrown:
That's my bad!!! I didn't copy over the presets in the karma.config!!

deathbearbrown:
You should not see that syntax error any more

deathbearbrown:
I added a little test for primative/Guide.js to make sure everything is cool. I will work on adding new tests every morning until we are caught up. :)

PR #273, opened 2016-03-03
Implement inheritance utility method

kadamwhite:

This standardizes the inheritance scheme used by constructors in a more robust manner

- Move existing utilities file into a directory
- Add and consume a new inheritance utility module

#### This PR includes:
- [ ] Tests
- [x] functional comments
- `n/a` high level description of any new components
- Regression

@deathbearbrown I could use a jsdoc sanity-check too, not 100% on how to document modules like this

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "eslint-loader": "^1.2.1",
     "eslint-plugin-react": "^3.16.1",
     "fs": "0.0.2",
+    "inherits": "^2.0.1",
     "istanbul": "^0.3.21",
     "jshint": "^2.8.0",
     "json-loader": "^0.5.4",
--- a/src/js/model/primitives/Guide.js
+++ b/src/js/model/primitives/Guide.js
@@ -1,4 +1,5 @@
 var Primitive = require('./Primitive'),
+    inherits = require('inherits'),
     model = require('../'),
     lookup = model.primitive;
 
@@ -57,8 +58,7 @@ function Guide(gtype, type, scale) {
   return Primitive.call(this);
 }
 
-Guide.prototype = Object.create(Primitive.prototype);
-Guide.prototype.constructor = Guide;
+inherits(Guide, Primitive);
 
 // TODO: Map guide properties to signals.
 Guide.prototype.init = function() {
--- a/src/js/model/primitives/Scale.js
+++ b/src/js/model/primitives/Scale.js
@@ -1,6 +1,7 @@
 var dl = require('datalib'),
     Primitive = require('./Primitive'),
     model = require('../'),
+    inherits = require('inherits'),
     lookup = model.primitive,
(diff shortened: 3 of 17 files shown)


arvind:
I think it would be good to switch _all_ files to use this pattern as part of this PR? `src/primitives/data/`, `src/primitives/Guide`, `src/primitives/Scale` and `src/transforms` are pending. 

I also wonder whether this might be an over-optimization? This PR requires add a long `require` string to each file, which will not get minified/uglified (a similar problem added a non-trivial amount to our minified file size for Vega). More problematic, IMHO, is the long relative path to get up to the `util` library, which strikes me as more error-prone than having an insitu `Object.create` call as we do now.

kadamwhite:
The repetition of that string should be negligible after gzipping the library, but there's other ways we could work around that problem too. I do not see it as an per optimization because it permits us to take a comparatively important piece of functionality and abstract it into a testable module that can be reliably counted upon to act in a predictable manner; object creation is concise but repetitious, and repetition is error prone. 

In terms of the specific inheritance mechanic that came out of a conversation with @gnarf so I will ask them to weigh in on this concern as well

gnarf:
The webpack built output would never have the require strings... It ends up converting it to something like `var inherits = __webpack_require(5)`

arvind:
Thanks, good to know! My larger concern was with the difficulty of specifying the correct set of `../../` to up-level to the util directory but I guess that's something that can be caught at build-time, is easily fixable, and likely worth the tradeoff for ensuring that inheritance always works correctly. Happy to merge this once all the other OO classes have been switched to this new pattern.

gnarf:
I also think you might just want to use [`require('inherits')`](https://github.com/isaacs/inherits) -- the code for it seems pretty exactly what you wrote here anyway: https://github.com/isaacs/inherits/blob/master/inherits_browser.js

It uses `require('util').inherits` in node environments, and writes it's own in browser envrionments

kadamwhite:
This PR's been updated to utilize the inherits module per @gnarf's suggestion, and I've applied that everywhere I could see this inheritance pattern being used. @arvind ready for test and review.

kadamwhite:
> The webpack built output would never have the require strings... It ends up converting it to something like `var inherits = __webpack_require(5)`

Thanks for the reminder, @gnarf ! That's great to know

arvind:
Thanks @kadamwhite! Minor stylistic point, I like to do external requires first and then internal ones, so I moved the `inherits` require in a [few places](https://github.com/vega/lyra/compare/84e12f9...edc840d)+tested.

kadamwhite:
@arvind I prefer that too, and shame on me for letting it slide in deference to increasing the speed I could copy-paste :D Thanks for cleaning that up, I'll keep an eye on that convention in the future.

PR #277, opened 2016-03-07
Clean up application initialization code

kadamwhite:

The index.js file which kicks off the entire application leaked a number of
global variables and defined functions that were not used anywhere else in
the codebase, likely as the remnant of earlier debugging code.

- Remove unused `f`, `f2` and `f3` functions
- Make variable declarations local
- Do not store references to the rect and symbol created for demo purposes
- Remove commented-out code

- Basic regression/smoke test
- @arvind's assurance that these methods aren't being used globally: I
  couldn't find them with a code search but verification would be useful

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -14,14 +14,14 @@ model.init();
 // Initialize components
 Sidebars = require('./components');
 
-g = model.Scene.child('marks.group');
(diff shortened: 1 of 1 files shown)


arvind:
Thanks @kadamwhite!

PR #278, opened 2016-03-07
Standardize curly brace usage across project

kadamwhite:

I'd written this patch last week and forgot to submit it; it rebased cleanly so here goes: this standardizes all conditional blocks to use curly braces, even for single-line `if` statements.

My recollection is that this rule was contentious, but my personal feeling is that the increased consistency (all blocks are curly'd) improves code scanning speed and reduces the risk for statements to escape the conditional block.

- Use curly braces everywhere (`curly` rule in eslint)
- Always follow a closing brace with a newline (`brace-style` rule in eslint)

- Smoke test to validate that no regressions were introduced
- `npm run lint | grep curly` to validate that no curly-brace inconsistencies are found

### DIFF
--- a/eslint/eslint-defaults.yml
+++ b/eslint/eslint-defaults.yml
@@ -136,8 +136,8 @@ rules:
     - never
   brace-style:
     - 2
-    - stroustrup
-    - allowSingleLine: true
+    - 1tbs
+    - allowSingleLine: false
   camelcase:
     - 2
     - properties: never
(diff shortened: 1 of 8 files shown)


arvind:
I agree with standardizing on curly brace style, and replacing single-line `if`s with a block too. However, I would prefer to use the [1tbs brace-style](http://eslint.org/docs/rules/brace-style.html#tbs) -- i.e., a new-line only after the _final_ closing brace of a block, rather than every one.

kadamwhite:
@arvind I'm fine with 1tbs; in the interest of focusing on implementation, the way I would like to proceed would be to add a commit changing to 1tbs, then merge without any further code changes—I'll revisit 1tbs conformity later on when I need something mindless to do while my brain works through a problem. Sound good?

arvind:
Yup, makes sense to me! Thanks @kadamwhite!

Issue #227, opened 2016-02-14
Support text marks


arvind:
Lyra 2 should support Vega's `text` mark type. This will involve creating:
- [x] an primitive mark class.
- [x] a manipulators transform class
- [x] property inspector
- [ ] test cases for rules

kadamwhite:
First three points ticked off in #281

kadamwhite:
@arvind can you clarify what "test cases for rules" means -- functional test cases, or unit tests?

arvind:
I meant functional test cases. I had started on some for [symbol marks](https://github.com/vega/lyra/blob/lyra2/test/rules.test.js) which I used to [produce PNGs](https://github.com/vega/lyra/blob/lyra2/test/resources/rules-fill.png). Producing PNGs affords visual inspection, but I'm not sure how robust it is against variations in server-side rendering (e.g., fonts). Another alternative is to generate the SVG strings and compare. Both of these, though, could come later and for now, we can just verify that rules produce the necessary properties on the model.

kadamwhite:
@arvind thanks for clarifying. I am going to close this now, then, and we can open more specific issues for iteration on text mark support/testing/etc as they occur to us.

PR #281, opened 2016-03-09
Implement text marks

kadamwhite:

Add support for Text marks and initialize the scene with a dummy text mark instance

- Add Text mark with associated manipulators and properties panel. Supported properties should have parity with lyra 1.
- Permit properties window to scroll vertically when it doesn't fit on my teeny laptop screen (very rough impl)

#### This PR includes:
- [x] Tests
- [x] functional comments
### To-Do:
- [x] define connectors
- [x] define channels
- [x] tests!
- [ ] ~~Move hard-coded option dropdowns into vega schema~~

LYRA-226

For #227

### DIFF
--- a/src/js/components/inspectors/Text.jsx
+++ b/src/js/components/inspectors/Text.jsx
@@ -23,11 +23,18 @@ var Text = React.createClass({
 
         <h3>Font</h3>
 
+        <Property name="font" label="Style"
+          primitive={primitive}
+          type="select"
+          opts={['Helvetica', 'Verdana', 'Georgia', 'Palatino', 'Garamond', 'Trebuchet MS']}
+          canDrop={true}
+          field={update.font.field}
+          signal={update.font.signal} />
+
         <Property name="fontSize" label="Size"
           type="number"
           primitive={primitive}
           canDrop={true}
-          scale={update.fontSize.scale}
           field={update.fontSize.field}
           signal={update.fontSize.signal} />
 
@@ -36,7 +43,6 @@ var Text = React.createClass({
           type="select"
           opts={['normal', 'bold']}
           canDrop={true}
-          scale={update.fontWeight.scale}
           field={update.fontWeight.field}
           signal={update.fontWeight.signal} />
 
@@ -45,7 +51,6 @@ var Text = React.createClass({
           type="select"
           opts={['normal', 'italic']}
           canDrop={true}
-          scale={update.fontStyle.scale}
           field={update.fontStyle.field}
           signal={update.fontStyle.signal} />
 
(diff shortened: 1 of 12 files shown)


kadamwhite:
text smoke

arvind:
One other direct-manipulation interaction that would be nice: being able to rotate the text when the command-key is pressed and you drag around a handle. This sort of interaction is present in Keynote, Omnigraffle, etc. for example.

kadamwhite:
Per discussion with @arvind in Slack, hard-coding options for alignment, font etc in the property inspector components is sufficient for now. I will open an issue to document moving to json schema.

kadamwhite:
@arvind what do you think about doing it through an additional handle? (not sure if it's possible to customize that cursor to be something other than a down-arrow though)
text rotation
I suggest this because I'm not aware of any other primitives we currently support having alternate handle-based manipulation modes given modifier keys, so I'm concerned about discoverability

src/js/components/inspectors/Text.jsx
```diff
@@ -0,0 +1,133 @@
...
+          field={update.text.field}
+          scale={update.text.scale}
+          signal={update.text.signal} />
+
+        <h3>Font</h3>
+
+        <Property name="font" label="Style"
+          primitive={primitive}
+          type="select"
+          opts={['Helvetica', 'Verdana', 'Georgia', 'Palatino', 'Garamond', 'Trebuchet MS']}
```
deathbearbrown: I don't know if this matters, but symbol mark contains all it's special info where the mark is defined. 
https://github.com/vega/lyra/blob/lyra2/src/js/model/primitives/marks/Symbol.js#L56

called in the inspector here: https://github.com/vega/lyra/blob/lyra2/src/js/components/inspectors/Symbol.jsx#L35

kadamwhite: That's not a bad pattern, I'll switch to that to keep the domain of these controls out of the component code. Thanks, @deathbearbrown !

kadamwhite: @arvind added an issue for the JSON schema q over in #284

src/scss/inspector.scss
```diff
@@ -5,6 +5,7 @@
   bottom: 0;
   width: $inspector-width;
   background-color: $lblue-bg;
+  overflow-y: scroll;
```
deathbearbrown, line 8: +1 lol

deathbearbrown:
so far what I see here looks good.

deathbearbrown:
I know what that console warning is:
https://github.com/vega/lyra/blob/lyra2/src/js/model/rules/VLSingle.js#L10
You need to add `text: 'text'` here

kadamwhite:
@arvind @deathbearbrown This one's in your court to review/merge; ping me in slack if any further changes are needed. Thanks for tracking down that bug, Sue!

deathbearbrown:
+1 from me, I functionally tested and reviewed the code yesterday.  It just needed that array updated to work perfect, which you did :)

PR #285, opened 2016-03-09
eslint compliants for .test.js specs

kadamwhite:
This PR addresses most eslint errors present in `*.test.js` mocha spec files.

The one remaining issue is the use of `undefined`, which is currently flagged as an error in eslint; I tend to prefer `typeof x === 'undefined'` when doing equality checks, but in this case `undefined` is actually standing in for a missing argument to a function so there's no comparably-obvious way to represent the usage than what's there already. For that reason I left it be until we could discuss.

To test, run `node_modules/.bin/eslint --ext .test.js src` to invoke eslint against just the .test.js spec files.

### DIFF
--- a/src/js/model/primitives/Guide.test.js
+++ b/src/js/model/primitives/Guide.test.js
@@ -1,4 +1,5 @@
 /* eslint no-unused-expressions:0 */
+'use strict';
 var expect = require('chai').expect;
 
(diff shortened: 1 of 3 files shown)


deathbearbrown:
+1 waiting for travis

PR #286, opened 2016-03-09
Add an "eslint:branch" npm script to lint only changed files

kadamwhite:
This script is a stop-gap until we get to passing eslint (at which point we'd want to integrate eslint as a pre-task for `npm test`) -- it uses git's CLI to determine the commit at which the working branch diverged from our default `lyra2` branch, then runs only those files through eslint. I submit it as an easy way for us to check that we're conforming to our style rules before we PR a new branch, since going through the aggregated eslint output is a bit of a pain.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   },
   "scripts": {
     "lint": "eslint --ext .js,.jsx src",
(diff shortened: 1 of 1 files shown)


deathbearbrown:
+1, this is a very very very good idea

PR #287, opened 2016-03-09
getName wasn't returning the constructor name

deathbearbrown:
The problem:
I made a getter for the wrong thing on the inspector.  It should have been getting the name of the constructor, not the name of the primitive.  This broke the inspector.  I'm a dope.

### DIFF
--- a/src/js/components/Inspector.jsx
+++ b/src/js/components/Inspector.jsx
@@ -12,7 +12,7 @@ var Inspector = React.createClass({
     var props = this.props,
         primitive = lookup(props.id),
         from = lookup(primitive.from),
-        ctor = primitive.getName(),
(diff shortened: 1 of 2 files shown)


src/js/components/Inspector.jsx
```diff
@@ -12,7 +12,7 @@ var Inspector = React.createClass({
     var props = this.props,
         primitive = lookup(props.id),
         from = lookup(primitive.from),
-        ctor = primitive.getName(),
+        ctor = primitive.constructor.name,
```
kadamwhite, line 15: Wouldn't this need to be primitive.prototype.constructor.name...? I'm not sure I understand the nuance of what this is doing.

kadamwhite:
Reproduced and this fixes the bug.

PR #290, opened 2016-03-10
Tests/support jsx and generate coverage

kadamwhite:

We have no reporting around the amount of test coverage within Lyra. This PR utilizes [karma-coverage](https://github.com/karma-runner/karma-coverage) to provide that reporting.

- Permit test specs to be saved as `.test.jsx`, instead of solely `test.js`
- Add Chrome launcher to karma config and set `npm run test:dev` to run in captured Chrome instance by default, to permit in-browser debugging
- Use the ["alternative" karma-webpack loading strategy](https://github.com/webpack/karma-webpack#alternative-usage) to pull in test specs, for performance and to permit loading files not yet under test for more accurate coverage reporting
- Implement code coverage reporting using karma-coverage
#### This PR includes:
- [x] functional comments
- `npm install`
- `npm test`
- `npm start`
- Browse to http://localhost:8080/coverage to view coverage reports (one report is generated for each browser in which Karma runs the tests, to expose browser inconsistencies).

Example coverage report at this point in time:

image
### IMPORTANT NOTE:

"Coverage" is determined based on "does this code execute during any unit test." That is a simplistic model that does not account for whether or not the unit test in question actually tests _that code_. As an example, the Utilities file is currently rated as having very high coverage, despite the fact that there are no tests specifically for those utility methods. This happens because those utilities are called from within other code that _is_ being tested, such as the Rect mark and its attendant transforms.

The takeaway is that coverage on its own is not a perfect metric; we need to have awareness of which areas of the codebase are being specifically exercised, and which are being tested only incidentally.


deathbearbrown:
Testing this, I can't see the bar charts for the coverage page.  I get a warning for an unmet dependency 'watchify'

I am going to throw out my node_modules folder and do a fresh install so I know it's not my local env.

kadamwhite:
I'll do a clean reset too, to see

deathbearbrown:
Functional +1 after derpin around

deathbearbrown:
+1 to code.  I will merge.  

Asked my question offline, so I'll just put it here for sake of history.  It doesn't matter if the extension of the file (eg .jsx) matches the test file.

PR #293, opened 2016-03-11
Add NPM test scripts for different browser/OS environments

kadamwhite:
**Description of the problem this pull request solves**

A tool is only as good as its supported environments: we should be testing
Lyra in as many browsers as possible.

For development purposes, there are now two watch-mode karma script aliases:
- `npm run test:dev` will spin up a karma watcher that runs tests in PhantomJS
- `npm run test:chrome` will do the same but use a captured Chrome instance,
  which can be useful for debugging purposes

For acceptance testing, we now have two platform-specific aliases that run
Karma against all the browsers that a developer might reasonably have installed
on their development OS of choice:
- `npm run test:osx` runs PhantomJS, Chrome, Firefox, and Safari
- `npm run test:win` runs PhantomJS, Chrome, Firefox, and IE


deathbearbrown:
Functional test +1 for osx.  I have to update everything on my windows machine tho to functional test  `npm run test:win`

kadamwhite:
Just verified `npm test:win` in WIndows 10, and added [custom launchers](https://github.com/karma-runner/karma-ie-launcher#configuration) for IE9 and IE10 emulation modes (not sure how accurate they are). At present there is no official karma launcher for Edge.

deathbearbrown:
Going to merge this and if anything weird comes up with win later, we can fix it

PR #298, opened 2016-03-14
Knock out about half of the remaining eslint errors

kadamwhite:

This PR reduces the noise of running `npm run lint` or `lint:branch` by cleaning up several broad categories of lint errors: JSDoc return statements, curly brace usage and block indentation, no need to use an `else` if all prior branches have `return`'d, etc.

It also makes the exposure of `model` and `Sidebar` in the global scope an explicit action.
- `npm test`
- `npm run lint` should return ~86 issues, rather than over 175 in lyra2 branch
- Smoke

### DIFF
--- a/karma.config.js
+++ b/karma.config.js
@@ -1,3 +1,6 @@
+/* eslint-env node */
+/* eslint quote-props:0 */
+'use strict';
 var argv = require('yargs').argv;
 var path = require('path');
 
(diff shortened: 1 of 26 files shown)


kadamwhite:
a8be12c really ought to say "almost" all files... :P I look forward to figuring out why adding `use strict` to the scales rule breaks area mark channel drop actions

kadamwhite:
Merging to have a clean slate for a new branch for migrating model components to a redux store; if there's any bugs open an issue and hit me with it :)

PR #299, opened 2016-03-16
Add redux dependencies and separate primitive getter from setter

kadamwhite:

Redux's unidirectional data flow means that the store only exposes getters, while "setters" are replaced by dispatched actions. To make tracing the behavior of the model setters and getters easier (and therefore to make switching the logic out to use a redux store easier), we separate half of `model.primitive` into a new `model.lookup` function.

- Add redux, immutable, and redux dev tools to package
- Extract primitive lookup code into a `model.lookup` method
- `npm test`
- Smoke


deathbearbrown:
+1 code and functional.  I'm mergin

PR #300, opened 2016-03-16
WIP: redux boilerplate set up

deathbearbrown:

Set up redux store for the project. 

- Adds redux, redux-react, immutable.js and  Webpack hot module replacement for reducers to package.json
- create a store
- add a combiner for reducers
- sample reducer for a todo app, not to be used for us, but to show what it looks like
- add provider to index.jsx

**TODO:**   
- [x] make store immutable 
- [x] linting (lol)
- [ ] tests

other notes:

I am using this convention:

```
var foo = function(){};
module.export = foo
```

instead of 

```
module.export = function(){};
```

because it makes it easier for me to step through code when a require fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "react": "^0.14.3",
     "react-addons-test-utils": "^0.14.6",
     "react-dom": "^0.14.3",
+    "react-redux": "^4.4.1",
     "redux": "^3.3.1",
     "redux-devtools": "^3.1.1",
     "sinon": "^1.17.2",
@@ -97,7 +98,6 @@
     "start": "webpack-dev-server --devtool=cheap-module-source-map --inline"
   },
   "dependencies": {
-    "jsdoc": "^3.4.0",
-    "react-redux": "^4.4.1"
+    "jsdoc": "^3.4.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "react-redux": "^4.4.1",
     "redux": "^3.3.1",
     "redux-devtools": "^3.1.1",
+    "redux-immutable": "^3.0.6",
     "sinon": "^1.17.2",
     "sinon-chai": "^2.8.0",
     "string.prototype.startswith": "^0.2.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -94,6 +97,7 @@
     "start": "webpack-dev-server --devtool=cheap-module-source-map --inline"
   },
   "dependencies": {
-    "jsdoc": "^3.4.0"
+    "jsdoc": "^3.4.0",
+    "react-redux": "^4.4.1"
```
kadamwhite: I realize I don't know whether there's a significant reason to use `dependencies` over `devDependencies`.

src/js/store/configureStore.js
```diff
@@ -0,0 +1,16 @@
...
+var rootReducer = require('../reducers');
+
+var configureStore = function(initialState) {
+  var store = createStore(rootReducer, initialState);
+  if (module.hot) {
+    // Enable Webpack hot module replacement for reducers
+    module.hot.accept('../reducers', () => {
+      var nextReducer = require('../reducers').default;
+      store.replaceReducer(nextReducer)
+    })
```
kadamwhite: Part of me wants to avoid doing any of this until we are in a better position to utilize hot-swapping; my experience with discussing hot swap configurations with our colleagues suggests to me that it's just going to be noise for now

package.json
deathbearbrown: oh I fixed that, I forgot to push it before I left

src/js/store/configureStore.js
deathbearbrown: ok, I only noticed this convention in a couple examples.  we can add it later if we need it :) :+1:

kadamwhite:
Note: this PR breaks the ability to select a mark on the stage, which dovetails with work I've been doing to port that functionality to a reducer in a branch I created off of this one.

deathbearbrown:
This has the immutable store set up.  Had to troubleshoot why combinedReducers would break.  I decided to use:
https://github.com/gajus/redux-immutable

There is another called redux-immutablejs (these names... ) but it doesn't configure in a way that is similar to the standard way I had before. That was annoying.  Anyway this is ok now.

kadamwhite:
:+1:. I know this is still WIP but I'm going to go ahead and merge it, to make the process of creating additional branches this morning more straightforward. Thanks for setting this up!

PR #301, opened 2016-03-18
Implement selected mark with redux

kadamwhite:

This PR re-implements mark selection using the Redux infrastructure introduced in #300. Rather than defining methods on individual components that interact with each other, Vega signals, and the model itself, we move towards using dispatched store actions to broadcast changes from components, and deduce component properties from the state of that store.

The current situation is not perfect, because it still relies on manually calling the vega signal to select the current mark; we should discuss ways to move this logic to the Redux store.
#### This PR includes:
- [ ] Tests
- [x] functional comments
- [?] high level description of any new components
- Smoke smoke smoke what did I break please find it and tell me


src/js/reducers/expandedLayers.js
```diff
@@ -0,0 +1,29 @@
...
+  if (typeof state === 'undefined') {
+    return Immutable.Map();
+  }
+  if (action.type === 'EXPAND_LAYERS') {
+    newState = action.layerIds.reduce(function(layers, layerId) {
+      layers[layerId] = true;
+      return layers;
+    }, {});
+    return assign({}, state, newState);
+  }
```
kadamwhite, line 15: The key thing I need help with is that I couldn't get these to work by using Immutable methods; the initial state should be an immutable map, but any subsequent change creates a werid, not-quite-Immutable object. `return state.merge()` did not work the way I expected.

kadamwhite:
This is now ready for final review. Tests will follow as time permits (I know, I know, I hate saying it too)

src/js/index.js
```diff
@@ -14,21 +14,23 @@ model.init();
...
-g.child('marks.line');
+g2.child('marks.line');
 g.child('marks.text');
 g.child('marks.area');
 
 Promise.all([
-  p._source.init({url: 'http://vega.github.io/vega-editor/app/data/cars.json'}),
-  p2._source.init({url: 'http://vega.github.io/vega-editor/app/data/jobs.json'}),
-  p3._source.init({url: 'http://vega.github.io/vega-editor/app/data/gapminder.json'})
+  p._source.init({url: '/data/cars.json'}),
```
deathbearbrown, line 31: did you do this intentionally or do you can test stuff on the train?

src/js/store/defaultState.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+var Immutable = require('immutable');
+
+// Create immutable state
+module.exports = Immutable.fromJS({
```
deathbearbrown, line 6: yus, I really like how you pulled out the default state into it's own file

src/js/components/Sidebars.jsx
```diff
@@ -4,83 +4,29 @@ var React = require('react'),
...
     LayerList = require('./LayerList'),
     Inspector = require('./Inspector'),
-    model = require('../model'),
-    sg = require('../model/signals'),
-    lookup = model.lookup;
+    model = require('../model');
 
 var Sidebars = React.createClass({
   getInitialState: function() {
+    window.sidebar = this;
```
deathbearbrown: where are we using window.sidebar?  Do we reference it anywhere else?

src/js/components/ScaleList.jsx
```diff
@@ -1,8 +1,22 @@
...
     ContentEditable = require('./ContentEditable');
 
-var ScaleList = React.createClass({
+function mapStateToProps(reduxState, ownProps) {
+  return {
+    selected: reduxState.get('selectedMark')
+  };
+}
+
+var ScaleList = connect(
```
deathbearbrown, line 12: I'd rather  we do `module.exports = connect(mapStateToProps,ScaleList);`  just because it's more readable.

Kadam I feel bad having opinions in code reviews.

src/js/index.js
kadamwhite, line 31: This was intentional so that I could test things on the train -- I mentioned it in slack, it's a sep commit i can rebase out.

kadamwhite, line 31: (and to clarify, I _will_ rebase that out, but I'm interested in adding this in anyway at least for a time because it dramatically speeds up the speed of reload on the app)

src/js/components/Sidebars.jsx
kadamwhite: Stray debugging code, thanks for the catch! Will remove.

src/js/components/ScaleList.jsx
kadamwhite, line 12: Please don't feel bad; I agree with your sentiment and I appreciate your raising it!

kadamwhite, line 12: I take it back, we can't actually make this change right now without further modularizing the code or making otherwise-unnecessary intermediate identifiers (LayerList for example has a non-exported, `connect`ed `Group` component that is used within the exported class). We _should_ fragment this, and I am noting this as a preferred code style, but I would like to keep this as-is for the purpose of this PR.

src/js/index.js
```diff
@@ -11,30 +11,32 @@ var model = require('./model');
 model.init();
 
 // Initialize components
-var Sidebars = require('./components');
+var ui = require('./components');
 
 var g = model.Scene.child('marks.group'),
+    g2 = g.child('marks.group'),
+    g3 = g2.child('marks.group'),
```
deathbearbrown, line 18: from the functional test:
When you add groups it shows up in the editor layers. Do we want the g3 group for rects to be nested under g2?  Wouldn't we want it all in one group like before?  Was this just test code to make sure things were working?

pipeline

src/js/components/ScaleList.jsx
deathbearbrown, line 12: ok!

src/js/index.js
kadamwhite, line 18: Correct, I added some nested groups when creating the primitives in `js/index.js`. Nesting adds a notable amount of complexity to the tree traversals involved in selection, so inasmuch as all of our homemade marks are for debugging/testing purposes, it felt appropriate to add some new levels to the hierarchy. If you object to them I can take them out but I have found them to be useful for smoke testing.

deathbearbrown:
I tried to make an animated gif of this, but chrome has decided to stop letting drag and drop work for some reason.

The pipelines inspector dropdown isn't updating after data is dragged onto a mark 

screen shot 2016-03-18 at 11 39 33 am

src/js/index.js
deathbearbrown, line 18: I think that's fine.  I just wanted to make sure it was intentional and not debugging code.

kadamwhite:
That pipelines dropdown appears to be broken in lyra2 as well, not sure when that happened

deathbearbrown:
I am +1 after review and functional test. I don't think pipelines is a blocker.

PR #305, opened 2016-03-18
Implement pure export functions

kadamwhite:

The current export logic is resource-based and leverages the model instance-ness of primitives. As we move to redux we will want to make the primitives be stored as immutable, raw objects rather than mutable instances; in order to anticipate this shift I have begun to convert the exporter logic into utility functions.

Currently I'm running into trouble where any primitive with signal values will not successfully pass through the primitive export method: there's an exception within the signal code relating to `sg.init`. PR'ing so @deathbearbrown can take a look.

To test, pull down branch and run the unit test suite.

### DIFF
--- a/src/js/model/primitives/Scale.js
+++ b/src/js/model/primitives/Scale.js
@@ -6,6 +6,18 @@ var dl = require('datalib'),
     lookup = model.lookup,
     names = {};
 
+// To prevent name collisions.
+function rename(name) {
+  if (!name) {
+    name = '';
+  }
+  var count = 0;
+  while (names[name]) {
+    name += ++count;
+  }
+  return (names[name] = 1, name);
+}
+
 /**
  * @classdesc A Lyra Scale Primitive.
  *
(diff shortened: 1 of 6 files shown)


kadamwhite:
This is now ready for review. The PR implements exporters for all defined marks. The functions aren't "pure" in that they still have to lookup primitives via the model that do not get passed in, but they have been separated from the prototypes.

Tests exist to confirm parity between results of `[Mark].prototype.export()` and `exportUtil([mark])`.

src/js/model/primitives/Scale.js
```diff
@@ -65,12 +68,13 @@ function fmap(f) {
 }
 
 function dataRef(ref) {
+  console.log(ref);
```
deathbearbrown, line 71: console.log

kadamwhite:
This should be more relevant once we finish migrating properties into the store; I will close it for now but let's keep the branch around.

PR #306, opened 2016-03-22
Refactor reducers to prepare for undo/redo

kadamwhite:

Inspector may not be the best name, but selected mark and expanded layers are related so this PR
- groups them together within the redux store
- combines their reducers (not sure how I feel about this)
- Adds a module to define actions 
- Removes sample code
- better handles initial state within inspector component


kadamwhite:
@deathbearbrown I see selected layer and expanded layers as being related, so this PR changes the store state to be

``` js
{
  inspector: {
    selected: 6,
    expandedLayers: {
      '4': true,
      '8': true
    }
  }
}
```

Extremely open to :bike: shedding this

src/js/components/LayerList.jsx
```diff
@@ -13,8 +14,8 @@ var React = require('react'),
 var MARGIN_LEFT = 10;
 
 function mapStateToProps(reduxState, ownProps) {
-  var selectedMarkId = reduxState.get('selectedMark'),
-      expandedLayers = reduxState.get('expandedLayers').toJS();
+  var selectedMarkId = getIn(reduxState, 'inspector.selected'),
+      expandedLayers = getIn(reduxState, 'inspector.expandedLayers').toJS();
```
kadamwhite, line 18: This utility I wrote makes handling nesting easier, @deathbearbrown it may be of use to you in the work you're doing

src/js/reducers/inspector.js
```diff
@@ -0,0 +1,38 @@
...
+
+  if (action.type === actions.SELECT_MARK) {
+    return state.set('selected', action.markId);
+  }
+
+  if (action.type === actions.EXPAND_LAYERS) {
+    return action.layerIds.reduce(function(newState, layerId) {
+      return setIn(newState, 'expandedLayers.' + layerId, true);
+    }, state);
+  }
```
kadamwhite, line 26: @deathbearbrown thank you for the run-down on `.setIn`, with that API and the wrapper I wrote for it this reducer finally looks like a reducer rather than a clumsy merge-r!

deathbearbrown, line 26: yaaaaay!  this is great and will be super helpful

src/js/util/immutable-utils.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+function pathToArr(pathStr) {
+  return pathStr.split('.');
+}
+
+function getIn(structure, pathStr) {
```
deathbearbrown, line 7: I would add comments here to define that the structure is an immutable object  (it is right? am I crazy?)

kadamwhite, line 7: Comments: good call.

deathbearbrown:
I'm going to merge this now.  I tested it locally and it doesn't break anything and I need dat code. We can add comments afterwards.

PR #307, opened 2016-03-22
WIP: Implement history queue for existing store values

kadamwhite:
For #17, we need a history queue. Redux provides a [standard convention for tracking history](https://github.com/reactjs/redux/blob/master/docs/recipes/ImplementingUndoHistory.md), which we leverage via the [redux-undo](https://github.com/omnidan/redux-undo) library. Changes to `getIn` make the history state invisible to components if all they need to do is get the present value of a given property.

To test, apply this diff or replicate these lines in the `store/index.js` file:

``` diff
diff --git a/src/js/store/index.js b/src/js/store/index.js
index d2458e2..ae356cd 100644
--- a/src/js/store/index.js
+++ b/src/js/store/index.js
@@ -8,3 +8,7 @@ var configureStore = require('./configureStore');
  * @type {Store}
  */
 module.exports = configureStore(require('./defaultState'));
+
+window.undo = require('redux-undo').ActionCreators.undo;
+window.redo = require('redux-undo').ActionCreators.redo;
+window.store = module.exports;
```

You can then make a change (such as toggling open several layers), then run `store.dispatch(undo())` in the console several times to step backwards through time.

**Remaining Issue:**

This is what is happening at present when I toggle open the topmost group:

image

Events are stacking up, and we need to step through to identify why multiple actions are firing at once. This is why you need to undo() many times to see any change in the UI. Once this is resolved, each action should be granularly undoable and we can build a better interface for the history.

LYRA-267


kadamwhite:
This is still WIP because it depends on untangling the communication between the store and the view; #308 was pre-work for that cleanup, which is in progress

arvind:
Superseded by #387.

PR #308, opened 2016-03-22
Begin de-abstracting the internals of the signal module & interface

kadamwhite:
The code paths within the Signal module are very layered and difficult to trace; this PR begins to peel apart the combined getter and setter and the `model.signal` wrapper method in favor of explicitly calling getter or setters exposed by the signals API and manually triggering view updates via a call to model rather than relying on an overloaded return value from the signal's `.view` and model's `.signal` methods.

The code as-is exacerbates an existing issue where dragging data onto channels may not take effect, so this probably needs more work before it is safe to merge. I am submitting it as an early PR so that @deathbearbrown can see my thinking.

(Another major change here is that it re-structures the signals module code to be less different from the other modules)

### DIFF
--- a/src/js/model/signals/index.js
+++ b/src/js/model/signals/index.js
@@ -8,12 +8,17 @@ function api() {
   return signals;
 };
 
+// Augment the signals API with properties like SELECTED that define the
+// strings used to identify and trigger a given signal
+dl.extend(api, defaults.signalNames);
+
 function ref(name) {
   return {signal: ns(name)};
 }
 
-function init(name, val) {
-  signals[name = ns(name)] = {
+api.init = function(name, val) {
+  name = ns(name);
+  signals[name] = {
     name: name,
     init: val,
     _idx: dl.keys(signals).length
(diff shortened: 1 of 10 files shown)


src/js/model/signals/index.js
```diff
@@ -6,75 +6,89 @@ var dl = require('datalib'),
...
+};
 
-function value(name, val) {
+api.getValue = function(name) {
   var model = require('../'),
+      // `view` is a vega runtime component; view.signal is a getter/setter
       view = model.view,
-      sg = signals[name = ns(name)],
-      set = arguments.length === 2;
+      signalObj = signals[ns(name)];
```
deathbearbrown, line 33: nitpick, but pull out ns(name) into a variable

kadamwhite, line 33: yeah i went back and forth on that, didn't want to make a variable we never used again but i see how it could be obfuscating. I'll update.

kadamwhite, line 33: yeah i went back and forth on that, didn't want to make a variable we never used again but i see how it could be obfuscating. I'll update.

deathbearbrown:
+1 Code review. +1 functional

PR #309, opened 2016-03-25
Webpack & scss

deathbearbrown:
I have to do a ton of CSS work and I need css watch, so I updated webpack to handle scss files.
By putting the scss stuff in webpack, we are now making one less request in the app. 

In this PR I have the updated webpack stuff & I have moved the JSX files about for my scaffold restructure.  I wanted to get this mini PR in first with these changes. 

To test this run `npm start` and make sure it's doing it's thing. The only difference you should see if a checkered BG which is my test that images are working when referenced in the scss.


src/js/components/Sidebars.jsx
```diff
@@ -1,11 +1,11 @@
...
-    PipelineList = require('./pipelines/PipelineList'),
-    ScaleList = require('./ScaleList'),
-    LayerList = require('./LayerList'),
-    Inspector = require('./Inspector'),
+    InspectorSidebar = require('./InspectorSidebar'),
+    VisualSidebar = require('./VisualSidebar'),
+    PipelinesSidebar = require('./PipelinesSidebar'),
     model = require('../model');
 
+// Splitting each sidebar into it's column
```
kadamwhite: nit: `its`

webpack.config.js
```diff
@@ -1,5 +1,8 @@
+/* eslint-disable */
```
kadamwhite, line 1: Why is this necessary?

deathbearbrown, line 1: because eslint was going crazy when I ran 'lint:branch'

kadamwhite:
We'll want to adapt this so that we can extract the CSS independently from the JS bundles at some point, but I'm +1 on what's here

webpack.config.js
kadamwhite, line 1: Huh, odd. Ok.

PR #310, opened 2016-03-25
Move signals to redux store

kadamwhite:
Will update description later, but PR'ing for now to get eyes on it: this moves all signals into the redux store, and (vebosely) synchronizes the view to that store on each update.


deathbearbrown:
Functional tested and nothing is broken.  I am logging the store in console and when I update values on a primitive I see the changes in the store. 

Looking through code now. :)

src/js/model/primitives/marks/Mark.js
```diff
@@ -66,8 +66,9 @@ Mark.prototype.init = function() {
 
   for (k in update) {
     p = update[k];
-    if (p.value !== undefined) {
-      update[k] = dl.extend(sg.init(propSg(this, k), p.value),
+    if (typeof p.value !== 'undefined') {
+      sg.init(propSg(this, k), p.value);
+      update[k] = dl.extend(sg.reference(propSg(this, k)),
```
deathbearbrown: this is so much to unpack here.  Can you put a comment around this block just to say that this is where the signals are initialized?

deathbearbrown:
Nothing jumped out at me with the code at first glance.

src/js/model/primitives/marks/Mark.js
kadamwhite: Yes, I can do that! Thanks, @deathbearbrown

kadamwhite:
@deathbearbrown cleaned up the code, added a couple unit tests for the action creators, and added that comment you requested. I believe this is ready for final review and merge when you have time.

src/js/components/mixins/Parse.jsx
```diff
@@ -1,29 +1,8 @@
...
-      // then walk down the rendered Lyra scene to find a corresponding item.
-      item = hierarchy.findInItemTree(sceneRoot, [primitive._id].concat(parentGroupIds));
-      // If an item was found, set the Lyra mode signal so that the handles appear.
-      // As noted above, this logic should probably not exist here!
-      if (item !== null) {
-        sg.value(sg.SELECTED, item);
-        model.update();
-      }
-    });
+    model.parse();
```
deathbearbrown, line 6: Question... do we need this now? Maybe for a later refactor if we aren't doing anything other than `model.parse()` we should get rid of this?

src/js/model/primitives/marks/Mark.js
```diff
@@ -99,7 +103,7 @@ Mark.prototype.dataset = function(id) {
     return this;
   }
 
-  this.from = undefined;
+  delete this.from;
```
deathbearbrown: I don't think you need to delete from.  
It's already set as undefined in the Mark function.  I guess you could remove it from there and then add the property in here?

kadamwhite: How I read this (and @arvind can confirm) is that this code is meant to unset `this.from` if the dataset method is called with arguments that cause it to fall through all the `else if`'s... so we either need to keep the `= undefined`, which feels janky b/c we try to avoid using undefined as a value, even tho it does work in strict mode; or delete the property, which feels more explicit and accomplishes the same end. Initializing it to undefined to begin with still feels weird, b/c just omitting it entirely ought to accomplish the same goals; but I fixed this lint error about use of `undefined` then decided I didn't want to get too deep in the weeds by addressing the other as well. I would revert this if you don't feel it's appropriate, but I don't think I'd do anything further to this method b/c this patch was just for lint cleanup and isn't related to the store or signals.

src/js/components/mixins/Parse.jsx
kadamwhite, line 6: We do need this for now, because a re-parse of the model is needed in order to have certain updates take effect. I absolutely support removing the parse mixin in favor of an action that triggers that re-parse down the road, it simply felt like it was out of scope for this PR.

PR #316, opened 2016-03-30
WIP: Add New Mark

kadamwhite:

This PR adds @deathbearbrown's implementation of a UI and underlying code to support adding new marks to the vega visualization (LYRA-228)

In-flight changes include clarifying the schedule by which the model updates based on user actions, to avoid situations where we are trying to set signals in vega that do not yet exist.


kadamwhite:
Actually I am going to close this PR and open it from a new branch, just to triply ensure I don't push anything that'll get in the way of @deathbearbrown

PR #317, opened 2016-03-30
Implement UI for adding new marks

kadamwhite:

This PR adds @deathbearbrown's implementation of a UI and underlying code to support adding new marks to the vega visualization (LYRA-228)

The model re-parse is wrapped in a new `CancellablePromise` construct, which allows us to throw away any Vega re-render request that is still in progress at the time a new request is received.

A single store listener now orchestrates all of the data-flow _from_ redux _to_ vega.

Tests are present, though not as many as I'd like. Submitting this as-is for review and merge consideration because it can serve as the basis of much subsequent work.

@deathbearbrown this now addresses the auto-select new mark question we discussed


kadamwhite:
Tracking issues to iron out
- [x] after initializing vega view the model.view.model().scene() is null inside vega streams code, when it is NOT null inside Lyra's model
- [x] A newly-added mark should be auto-selected when it is added, but there's a race b/w adding the mark and setting it as visible

src/js/model/index.js
```diff
@@ -331,26 +345,78 @@ function updateSelectedMarkInVega() {
...
+    if (model.view) {
+      // Clear out the outdated vega spec: iterate through all registered
+      // signal streams and remove their event listeners
+      model.view.destroy();
+      model.view = null;
+    }
+    store.dispatch(parseStart());
+    // Don't start a reparse more often than 60 times a second
+    debouncedInitiateReparse();
+    // initiateReparse();
```
kadamwhite, line 386: This is the component of this about which I am least satisfied: if we do not debounce this, then the pipelines sometimes fail to load. Why, is anybody's guess. Hopefully moving that sort of action into the store will fix this naturally later on.

src/js/model/index.js
```diff
@@ -331,26 +345,78 @@ function updateSelectedMarkInVega() {
...
+  }
+  return shouldReparse;
+}
+
+// This store listener handles all of the data-flow FROM the Redux store TO the
+// Vega view: very little code outside of this method should be interacting with
+// the view directly.
+store.subscribe(function() {
```
kadamwhite, line 394: This is now the sole store listener responsible for interacting with Vega, which I feel to be much more efficient than the parallel listeners (and corresponding proliferation of state-checking conditionals) we had before.

deathbearbrown:
OH BOY IT SELECTS THE NEW MARK!!!

src/js/model/index.js
deathbearbrown, line 394: I like this a lot.

deathbearbrown:
@kadamwhite do you have any gotchas that I should be looking for here?  Because I've been functional testing and other than the sidebar updating at root level thing which I have fixed in my delete branch, it looks pretty solid.

src/js/util/simple-cancellable-promise.test.js
```diff
@@ -0,0 +1,220 @@
...
+      prom.cancel();
+      return 'foo';
+    }).then(function(result) {
+      if (result === 'foo') {
+        return 'did not cancel';
+      }
+    })).to.eventually.equal('did not cancel');
+  });
+
+  it('does not swallow errors within the initial promise', function() {
```
deathbearbrown, line 209: This is my favorite test description

deathbearbrown:
+1 code.  Your comments are really helpful when walking through what's going on.   I think your solution to the issue is great.

PR #319, opened 2016-03-31
Avoid generic identifiers like "Base"; prefer specificity

kadamwhite:

No functionality changes here; this is the result of my review of the app hierarchy, focusing on clarifying the relationships between components by using less-generic identifiers ("Area" instead of "Base," and "AreaInspector" when we're in a file with the Area Mark _and_ Inspector, e.g.).

- Rename a bunch of things

- Does the app break...? it should not.


deathbearbrown:
+1

PR #320, opened 2016-03-31
Add an incrementing counter utility

kadamwhite:

There are a lot of implicit incrementing counters in Lyra: this makes them explicit and tested. As we move on to a world where we can load state from a file or localstorage or some such, we will need to add in the ability to seed these counters at a base starting value for each type; making a utility sets us up to make that change as well.

- Implement an incrementing counter utility
- Utilize that utility within mark and primitive code

**This PR includes:**
- [x] Tests
- [x] functional comments

- Smoke test
- Unit tests pass


deathbearbrown:
+1  code and functional. :)

Issue #314, opened 2016-03-30
Suppress CSS load error in Karma environment


kadamwhite:
![image](https://cloud.githubusercontent.com/assets/442115/14147590/1b7a8970-f66a-11e5-84cd-4d0a7c632129.png)

Possibly related to https://stackoverflow.com/questions/32683440/handle-webpack-css-imports-when-testing-with-mocha

PR #321, opened 2016-04-01
Improve CancellablePromise and alter Karma config to avoid initializing the app while running tests

kadamwhite:

- Modify the CancellablePromise to permit its executor function to inspect its cancellation state
- Build a 10ms timeout into the model parse method to permit intermediate cancellation without having to debounce the entire parse method
- Throttle the vega update function as a separate store listener

**This PR includes:**
- [x] Tests
- [x] functional comments

- Smoke
- Unit test suite
- Functional

A conceptual weakness of our CancellablePromise is that it did not give itself any internal means to inspect the state of the cancellation, precluding conditional actions inside the promise executor function.

One common scenario for which this utility was written is the Parse cycle: imagine the following scenario,
1. User adds a mark
2. Mark registers its signals
3. User alters mark's signal values

The first two of these actions require the Vega model to be re-created; the third requires the first two to have taken place, or the signals will not be present when Lyra tries to set those signal values. Because mark addition and signal registration both require a Vega re-parse, we initialize the re-parse for each action; but we only care about the final state, not the state after (1) alone. Exposing promise cancellation status with a bound executor function permits the promise to be created with a built-in debounce of 10ms, so that if the 2nd action happens immediately after the first in execution order, we don't have to begin to invoke the model parse at all. This is slighty more efficient than simply throwing away the result of the parse.

The output of these debug statements:

``` diff
diff --git a/src/js/model/index.js b/src/js/model/index.js
index 9764276..c4b093a 100644
--- a/src/js/model/index.js
+++ b/src/js/model/index.js
@@ -229,6 +229,7 @@ var parsePromise = null;
  * parsed and rendered.
  */
 model.parse = function(el) {
+  console.log('parsing');
   el = el || '#vis';
   if (parsePromise) {
     // A parse is already in progress; cancel that parse's callbacks
@@ -237,15 +238,19 @@ model.parse = function(el) {

   // Start the newly-requested parse within a cancellable promise
   parsePromise = new CancellablePromise(function(resolve, reject) {
+    console.log('- Starting promise');

     // Debounce parse initiation very slightly to handle re-starts on subsequent
     // store listener digest cycles: CancellablePromise exposes its state through
     // this.cancel.
     var that = this;
     setTimeout(function() {
+      console.log('  - Checking for permission to continue...');
       if (that.cancelled) {
+      console.log('    - cancelled...');
         return;
       }
+      console.log('    - ok!');
       // Recreate the vega spec
       vg.parse.spec(model.manipulators(), function(err, chart) {
         if (err) {
@@ -257,6 +262,7 @@ model.parse = function(el) {
   });

   return parsePromise.then(function(chart) {
+    console.log('      - Done!');
     model.view = chart({
       el: el
     });
@@ -373,6 +379,7 @@ function updateAllSignals(state) {

 function initiateReparse() {
   model.parse().then(function() {
+    console.log('        - done parsing');
     store.dispatch(parseInProgress(false));
   });
 }
```

is this log:
image
You can see that of the 7 vega model re-parse requests received, this change causes only two vega parse actions to run, and only one parse callback to fire: the first five requests are cancelled within their 10ms debounce window, while the 6th is cancelled once the parse is in progress but before the chart has completed and been injected into the page.

This change allows us to remove the debounce on the parse method, cleaning up the model code a bit. We also move to using a throttled vega update method to avoid unnecessary updates.


kadamwhite:
I have hijacked this PR to solve a set of problems I encountered while testing this out:
1. Addresses #314 by using the raw-loader webpack loader to ignore SCSS content in the test environment,
2. Use the Webpack IgnorePlugin to prevent execution of the app entrypoint `js/index.js` file, and adapt the `test_index.js` file to not filter out all index files (because most of them get loaded anyway by the modules under test), but to only exclude the app and jsx entrypoints,
3. To prevent side-effects while testing store-related modules, pull all store listeners out into their own file and create them using dependency injection so that mocks can easily be passed in to test without requiring any interaction with singleton objects or the require cache
4. I have removed the `util/schema` utility, which has not been used at all since we joined the project. @arvind please let me know if this is inadvisable

deathbearbrown:
Ok I have functional stress tested the heck out of this and it's doing great.

src/js/index.js
```diff
@@ -16,6 +18,25 @@ var reparse = require('./actions/vegaInvalidate');
...
 model.init();
 
+// Set up the listeners that connect the model to the store
+var listeners = require('./store/listeners');
+
+// Bind the listener that will flow changes from the redux store into Vega
+store.subscribe(listeners.createStoreListener(
+  store,
+  model,
+  listeners.recreateVegaIfNecessary,
```
deathbearbrown: are these new instances or could you reference these methods from inside the file instead of passing them?

kadamwhite: They're just functions, and absolutely could be referenced from inside the file -- doing things this way is unorthodox but is intended to make it easier to test these listener methods, since we can pass in a mock store and model with spies to validate the branching. Those tests are yet to be written but my notion with the file was, all non-utility dependencies should be explicitly injected as arguments rather than relying on implicit relationships.

src/js/store/listeners.js
```diff
@@ -0,0 +1,177 @@
...
+ * the repostory and adopts a dependency-injection-based approach. The individual
+ * functions that handle certain types of Store -> Vega updates each take an
+ * argument representing the relevant store state, and an argument representing
+ * the Vega view (or the Lyra model itself): in this way their behavior can be
+ * validated using mock objects without having to set up a global store in the
+ * test environment.
+ *
+ * As for the primary `syncStoreToVega` listener, it is created through a function
+ * which takes in the store itself and all of the functions which it must orchestrate,
+ * and returns the final listener method.
```
kadamwhite, line 15: @deathbearbrown This comment block explains the rationale further

deathbearbrown:
is good +1 :dancers:

PR #323, opened 2016-04-01
Permit mark properties to be passed in on creation

kadamwhite:

Mark properties used to be set within the constructor in such a way that they could not be modified or overwritten until the object was fully instantiated and returned. That is inefficient for us because we want to be able to create primitives on demand from exported JSON or stored redux state. This PR alters the flow to allow each primitive to be instantiated by passing in a configuration object defining its state.

Marks to be updated to follow this pattern:
- [x] Area.js
- [x] Group.js
- [x] Line.js
- [x] Mark.js
- [x] Rect.js
- [ ] Scene.js
- [x] Symbol.js
- [x] Text.js

**This PR includes:**
- [x] Tests
- [ ] functional comments

- Smoke & functional
- Does anything break that used to work?

### DIFF
--- a/src/js/model/primitives/Primitive.js
+++ b/src/js/model/primitives/Primitive.js
@@ -19,9 +19,8 @@ var dl = require('datalib'),
  * @constructor
  */
 function Primitive() {
-  this._id = counter.global();
+  this._id = this._id || counter.global();
   model.primitive(this._id, this);
-  return this;
 }
 
 function _clean(spec, clean) {
--- a/src/js/model/primitives/Scale.js
+++ b/src/js/model/primitives/Scale.js
@@ -6,6 +6,15 @@ var dl = require('datalib'),
     lookup = model.lookup,
     names = {};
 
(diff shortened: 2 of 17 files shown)


kadamwhite:
This PR contains #321, so that one should be reviewed and merged first and then we can rebase this

src/js/util/schema.js
```diff
@@ -1,13 +0,0 @@
-'use strict';
-var vg = require('vega'),
-    vgSchema;
-
-/**
- * Return the Vega JSON schema, fetching it via XHR if it is not yet available
- * @returns {Object} The Vega JSON schema.
- */
-module.exports = function() {
```
deathbearbrown: Where was this even used?

kadamwhite: It wasn't, I realized that even when we started it wasn't being utilized

deathbearbrown:
Ok I have functional tested this before and after the rebase. 

The code looks good

PR #324, opened 2016-04-04
Avoid scene references

kadamwhite:

We are rapidly approaching (as in, next couple of hours) a world where a single, consistent reference to `model.Scene` no longer exists; the scene will be instantiated based on a raw object inside the store, and therefore the scene reference itself might be changed as the primitive instances are recreated or updated. To mitigate the impact of this on the rest of the codebase, this PR removes a couple straggling areas where the scene was being referenced directly; in reality none of these elements actually needed to get the scene's ID, since selection had been moved to the redux store for everything but pipeline and pipelines don't care about the scene. (But I did go ahead and move pipeline selection into the store anyway, just to be tidy).

- Control pipeline selection through redux
- Remove unneeded references to `model.Scene`


deathbearbrown:
Ok I have checked this out and rebased lyra2 into is. I get the Ranged error when I try to add data to a second axis.
screen shot 2016-04-05 at 3 24 25 pm

I am going to take a second and try to troubleshoot some of this. 

I do not get the errors when I test it on symbols.

deathbearbrown:
Clear all breaks. I think you know this is changing, I'm just commenting for the sake of history.

deathbearbrown:
All the marks seem to work except rect.  
**Steps to reproduce:**
- add a RECT mark.  
- Add data to an axis. (everything should be ok)
- Add data to the second axis (console errors)

deathbearbrown:
ok great now it stopped happening. >:C

deathbearbrown:
I am going to merge this and start an issue that has a bunch of functional testing flows, to make sure that the next PR is uber QA'd since so much stuff is in flight right now. :)

PR #326, opened 2016-04-05
React tooltips

deathbearbrown:
Adds React-tooltip to the project:
https://github.com/wwayne/react-tooltip
### Test

hover over trash icons on the sidebar, or + icons for Groups, Scales and Data.

Tooltips are only for informational text. Call to action hints will be located in a separate component.


kadamwhite:
tooltip
:+1:

Issue #318, opened 2016-03-30
Signals store implementation breaks direct action persistence


kadamwhite:
### Expected Behavior

When you interact with a mark via its handles, the mark's properties should persist once you select another mark.
### Actual Behavior

No direct manipulation-originated properties persist when you select another mark: it "snaps" back to where it was.
### Steps to reproduce behavior

![drag issue](https://cloud.githubusercontent.com/assets/442115/14160699/1f9d674c-f6ac-11e5-8385-88a4cdc09543.gif)

This happens because the SignalValue mixin does not set anything on the store. A simplistic fix would be to set store values via dispatched signals while dragging a mark, and to use handle interactions to set a flag that a direct manipulation is ongoing in order to suppress the "redux flows its signals back into vega" behavior while the user is still directly manipulating the mark.

(Introduced in #310)

kadamwhite:
@deathbearbrown & I discussed two potential approaches to this today, which I want to capture here to articulate my proposed solution. To recap, the bug exists because our current flows are as follows:

(flow 1): **Lyra-originated signal change** gets dispatched to **redux store**; a **store listener** passes that change _from_ the store on to **vega**.

(flow 2): An interaction within vega causes an **internal vega-originated signal change** which is listened to and rendered by the **properties component**, but it never makes it to the store.

What's happening in the above gif is that we're modifying a host of signal values at once by dragging the mark, but then when we use the property inspector to set one of them from Lyra's side that value gets added to the store and all of the other modified values are overwritten by their "stale" store equivalents, causing the mark to snap back.

The simplistic approach would be to always dispatch signal changes from the vega listeners that fire when a signal changes within vega, flowing those values into lyra's redux store. But the redux store listener tries to sync its state to vega, so what would happen here is that, say, you drag a mark to the right: its x property updates within vega and is dispatched to the redux store; meanwhile, you keep dragging. When the redux store learns about the signal it tries to set that in vega; the signal is now out of date, so your mark shakes back and forth between where vega thinks it should be going and where Lyra knows it was.

Solutions to that approach include
1. Dispatch the signal from the lyra vega listener, as usual, but flag it as originating from vega; something like the action `{ type: 'SET_SIGNAL', signal: 'some_signal', value: 600, fromVega: true }`. The store listener which syncs the lyra state back to Vega would then skip any values that are marked as having been last-touched by vega instead of lyra's React UI.
[...]

PR #327, opened 2016-04-05
Fix signals flow

kadamwhite:

As described in #318, signal changes originating from vega (e.g. drag and drop actions) were not making it back up to the redux store, meaning that when the user selects any other object after dragging a handle, the mark they modified "snaps" back to where it began.

Additionally, channel interactions only sporadically persisted; sometimes dropping a value onto a channel would fail with a warning that an error had occurred.

- sg.set _always_ flows changes both to the store _and_ to vega, if vega is available
- SignalValue mixin dispatches detected changes to the store to keep Redux in sync with Vega
- `.bind` function from the rules mixin has been renamed to `.rebind` to avoid confusion with `Function.prototype.bind`
- DataTable component has been updated so that the drag-and-drop action uses different state variables for data and for markup storage, fixing an issue where markup-targeted actions would null out the data pointer during a drag operation
- Rules code has been fixed to properly set the context of a function using `.call`
- DataTable code has been fixed to use array index for React `key` instead of `v._id`, because after a drop operation (droperation?) `v._id` becomes undefined. **This may point at a larger issue in how data is passed within the app** but fixing it was out of scope for this effort due to time constraint.

- #318 issue:
  - Add a mark, e.g. a symbol
  - Drag the symbol around the stage
  - Deselect the symbol by clicking the stage or adding another mark
  - Symbol should remain where you left it
- Channel issues
  - Drag and drop various data sources onto the channels of marks
  - Data should be bound as expected

### DIFF
--- a/src/js/components/pipelines/DataTable.jsx
+++ b/src/js/components/pipelines/DataTable.jsx
@@ -107,7 +107,7 @@ var DataTable = React.createClass({
         // vega-lite (see the rules index file) -- looks up what channel we're
         // on, finds a vega-lite property, puts that in the rules object,
         // calls vega lite compile, then iterates through each part of the rule.
-        prim.bind(cell.key, fullField._id);
+        prim.bindProp(cell.key, selectedField._id);
       }
     } catch (e) {
       console.warn('Unable to bind primitive');
--- a/src/js/components/pipelines/DataTable.jsx
+++ b/src/js/components/pipelines/DataTable.jsx
@@ -49,7 +49,12 @@ var DataTable = React.createClass({
         schema = this.props.dataset.schema();
 
     this.hideFull(evt);
-    this.setState({fullField: schema[name]});
+    this.setState({
+      // fullField is used for rendering
+      fullField: schema[name],
+      // selectedField is used to pass data around while dragging
+      selectedField: schema[name]
+    });
     this._fullField.style('display', 'block')
       .style('top', target.offsetTop);
   },
@@ -96,7 +101,7 @@ var DataTable = React.createClass({
   handleDragEnd: function(evt) {
     var sel = sg.get(sg.SELECTED),
         cell = sg.get(sg.CELL),
-        fullField = this.state.fullField,
+        selectedField = this.state.selectedField,
         dropped = sel._id && cell._id,
         prim;
 
--- a/src/js/components/pipelines/DataTable.jsx
+++ b/src/js/components/pipelines/DataTable.jsx
@@ -180,7 +180,7 @@ var DataTable = React.createClass({
                     onMouseOver={this.showFullField}>{k}</td>
                   {values.map(function(v, i) {
                     return (
-                      <td key={v._id} className={i % 2 ? 'even' : 'odd'}
+                      <td key={k + i} className={i % 2 ? 'even' : 'odd'}
                         onMouseOver={this.showFullValue}>{v[k]}</td>
                     );
                   }, this)}
--- a/src/js/model/rules/index.js
+++ b/src/js/model/rules/index.js
@@ -10,7 +10,7 @@ var dl = require('datalib'),
 
 function rules(prototype) {
 
-  prototype.bind = function(property, id, manual) {
+  prototype.bindProp = function(property, id, manual) {
     var rule = this._rule,
         from = this._from && lookup(this._from),
         obj = lookup(id),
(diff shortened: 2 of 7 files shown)


src/js/components/pipelines/DataTable.jsx
```diff
@@ -175,7 +180,7 @@ var DataTable = React.createClass({
                     onMouseOver={this.showFullField}>{k}</td>
                   {values.map(function(v, i) {
                     return (
-                      <td key={v._id} className={i % 2 ? 'even' : 'odd'}
+                      <td key={k + i} className={i % 2 ? 'even' : 'odd'}
```
kadamwhite, line 183: @arvind After dropping a field onto a channel, `v._id` is undefined on the next run. This fix suppresses the react error but I did not look into why `._id` is being unset.

src/js/model/rules/index.js
```diff
@@ -10,7 +10,7 @@ var dl = require('datalib'),
 
 function rules(prototype) {
 
-  prototype.bind = function(property, id, manual) {
+  prototype.bindProp = function(property, id, manual) {
```
kadamwhite, line 13: As a general rule I recommend against naming methods identically to important built-in constructs; this didn't overwrite anything b/c it was not a function, but it did confuse me for a while.

deathbearbrown:
for the sake of documentation, everything seems to be working when I functionally test, but when I drop data onto the mark I get this warning
`DOMPropertyOperations.js:168 The specified value "false" is not a valid number. The value must match to the following regular expression: -?(\d+|\d+\.\d+|\.\d+)([eE][-+]?\d+)?`

src/js/model/rules/scales.js
```diff
@@ -99,7 +99,7 @@ module.exports = function(parsed) {
     if (!def) {
       continue;
     }
-    if (!curr || !equals(def, curr)) {
+    if (!curr || !equals.call(this, def, curr)) {
```
kadamwhite, line 102: This was our sneaky sometimes-error for dropping fields: `.call` was used in all other cases, just not here. I'd recommend down the line to make this a function that takes its context as an argument to avoid excessive binding, `equals(this, def, curr)` e.g.

src/js/model/signals/index.js
```diff
@@ -92,9 +92,13 @@ api.get = function(name) {
...
   var model = require('../'),
       view = model.view;
+  // Always flow signals up to the store,
   if (!isDefault(name)) {
     store.dispatch(setSignal(name, val));
-  } else if (view && typeof view.signal === 'function') {
+  }
+
+  // and if we have a Vega view, flow signals down to Vega as well.
+  if (view && typeof view.signal === 'function') {
```
kadamwhite, line 101: This is what fixes #318

src/js/store/listeners.js
```diff
@@ -91,31 +91,6 @@ function updateSelectedMarkInVega(selectedMark, vegaView) {
...
- * Setting the same value on a signal twice should have no effect, so  we
- * naively set all signals on each store update to ensure store and Vega
- * stay in sync. Altering this to be smarter is one area to investigate should
- * any performance issues crop up later on, but signal writes are pretty fast.
- *
- * @param {Object} state - An Immutable state object
- * @param {Object} vegaView - A vega view instance to update with the selected mark
- * @returns {void}
- */
-function updateAllSignals(state, vegaView) {
```
kadamwhite, line 103: Since we now always update _both_ vega and the store with `sg.set`, we don't need this sync step anymore.

src/js/components/pipelines/DataTable.jsx
```diff
@@ -49,7 +49,12 @@ var DataTable = React.createClass({
         schema = this.props.dataset.schema();
 
     this.hideFull(evt);
-    this.setState({fullField: schema[name]});
+    this.setState({
+      // fullField is used for rendering
+      fullField: schema[name],
+      // selectedField is used to pass data around while dragging
+      selectedField: schema[name]
+    });
```
kadamwhite, line 57: Separating these out fixes the issue where `hideFull` would null out the data pointer needed by the drop handler

deathbearbrown:
Ok here's a bug and I need to make sure this doesn't exist in master, so take it with a grain of salt:

Steps:
- add a group
- add a symbol to that group
- add data to one axis of the symbol
- delete the symbol

You should see this error in console:
`ns.js:13 Uncaught TypeError: Cannot read property 'startsWith' of undefined`

src/js/model/rules/scales.js
deathbearbrown, line 102: this is a great catch

deathbearbrown:
haha ok so I manually applied the fix for master to this to test.  I have reviewed the code and I am +1

PR #330, opened 2016-04-05
use regex signal cleanup

deathbearbrown:
So I was playing with cleaning up the whole scene with scales, cause whatever and I notice that all the signals for a prim were not getting removed if you have bound data to an axis. 

So this will fix that, though I don't know if your refactor will change everything.

No tests because the test I would alter is skipped atm. :/

### DIFF
--- a/src/js/model/primitives/marks/Mark.js
+++ b/src/js/model/primitives/marks/Mark.js
@@ -216,10 +216,11 @@ Mark.prototype.remove = function() {
       key;
   for (key in update) {
     var signalName = update[key].signal;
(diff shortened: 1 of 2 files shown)


kadamwhite:
This is cool! I may subsume this branch into my WIP primitives-in-store one tomorrow, it would slot in nicely. Thanks, @deathbearbrown !

kadamwhite:
I have adopted this approach in my PR; if you're comfortable with my implementation I will close this one. Thank you for the direction!

deathbearbrown:
I'll close it :)

PR #332, opened 2016-04-07
Use store for all primitive properties

kadamwhite:
This moves primitive properties into the store, and uses the store to seed the `primitives` object in the model that is used for lookup.

`npm install` may be needed to test this.

Dispatching an action, e.g. PRIMITIVE_ADD_MARK, will be used to
1. create the primitive in the primitive store, with appropriate signal references
2. instantiate any signals in the signal store to match those references
3. set up streams in the signal store to handle mark interactions
### Key known issues:
- Channels are completely broken, because rules don't know how to update the store yet. I will need to pick @arvind's brain on the inner workings of rules to address this.
- Dragging a mark on the stage has gotten slower; one of the store updates is causing react to re-render more than it needs to or should.


kadamwhite:
vectors for investigating the selection issue:
- Manual inspection of `model.view.model().scene().items` between lyra2 and this branch shows the structure appears to be similar; but the logic in `updateSelectedMarkInVega` (`hierarchy.findInItemTree`) fails to find the mark.
- Diffing the JSON created from `model.manipulators()` between lyra2 and this branch shows that the only difference appears to be in ID numbers, not content, so it's unclear what could be causing the rendered vega graph to diverge.

deathbearbrown:
Heyas, @arvind found this https://github.com/vega/lyra/blob/lyra2/src/js/util/hierarchy.js#L102

When I modify that to

```
for (i = path.length - 1; i >= 0; --i) {

```

The selection works

kadamwhite:
Here's a formal TODO list for when this branch will be ready for review:
- [x] Can select a mark by clicking on it in the sidebar
- [x] Can delete an individual mark
- [x] Can "clear all"
- [ ] Can bind data to mark channels (this may take a while, not sure how best to proceed)
- [ ] Evaluate whether performance issue must be mitigated to proceed

kadamwhite:
@deathbearbrown mark instantiation, modification (excluding data binding) and removal (both individually and via "clear all") are now ready to test. deleteMark and clearScene actions are worth special note since they use redux-thunk and that's new for this project.

deathbearbrown:
I don't know how you want me to file these. I'm just going to put multiples in a comment for now so it's not too overwhelming. 

FIXED ~~FILL:   Initial color state is set to black.  Should be default blue.  It's important to have those match for the user to know what they are editing without being explicitly told~~

## screen shot 2016-04-12 at 10 49 10 am

FIXED ~~STROKE - the width isn't set, so if I change the color there is no update until I change the width. The property manager shows it as the middle of the graph instead of unset~~

## screen shot 2016-04-12 at 10 51 07 am

~~Can't change the shape of a symbol~~ FIXED

## screen shot 2016-04-12 at 10 53 56 am

Maybe bug, requires more discussion about expectations:
steps to see what I see:
1. create a group with a mark in it, create a mark outside that group
2. add data to axis on each mark
3. delete group
4. the scales still exist but the legends disappear

screen shot 2016-04-12 at 10 59 30 am

---

deathbearbrown:
Can't change a line's width
screen shot 2016-04-12 at 11 09 31 am
`Pulse already has a non-zero stamp.` this is the error I have  in console.   

I'm going to do a fresh checkout to make sure that I don't have any crummy stuff in this branch from the last time I checked it out. 

---

UPDATE:
~~I can still create the pulse errors.  I can change the stroke width, but I still get an error if I change the color.~~

---

UPDATE:  It's fixed!!! working beautifully!

deathbearbrown:
~~(OK on a clean branch.) I made a line graph. I changed the width of the stroke and it worked fine. I went to change the color and that didn't work and also the stroke property stopped working. Something wonky is happening so I'm gonna hold off until your next commit is in and then start again.~~
FIXED

kadamwhite:
I'm getting a lot of pulse errors when using the signal inspectors, in general. I'll investigate but any further specific examples you can find are helpful!

deathbearbrown:
@arvind what do these alignment properties for rect marks do?
screen shot 2016-04-12 at 11 42 00 am

arvind:
Re: Rectangle spatial properties, they are currently broken in `lyra2` as well. So likely not related to this PR, but we should investigate. Here's a GIF from an earlier commit demonstrating the intended behaviour. Basically, the horizontal and vertical position/sizing for rectangles is jointly determined by two properties: typically a start/end, a start/span, or a center/span. The drop downs determine which two properties are being used (and set `_disabled` on the others). 

rectspatial

deathbearbrown:
OK I have functional tested this like crazy.  All the marks are looking amazing.  I tested all properties. The only thing that isn't working is those drop downs on rect, but that's a known issue.

src/js/components/index.jsx
```diff
@@ -12,3 +12,4 @@ module.exports = ReactDOM.render(
   </Provider>,
   d3.select('#sidebars').node()
 );
+window.ui = module.exports;
```
deathbearbrown, line 15: we need to clean up these debuggy things in the app someday

src/js/model/index.js
```diff
@@ -24,16 +23,8 @@ var pipelines = [],
...
- * @returns {Object} The Lyra Model
- */
-model.init = function() {
-  var Scene = require('./primitives/marks/Scene');
-  model.Scene = new Scene().init();
-  store.dispatch(expandLayers([model.Scene._id]));
-  return this;
-};
+window.listeners = listeners;
+window.primitives = primitives;
```
deathbearbrown, line 27: <3 windows

kadamwhite, line 27: yeaaah one of my todos tonight is to audit all of the stray debugging code; i'm still usin' 'em tho so i it's OK for you can they live where they are through this merge? :D

src/js/model/primitives/marks/Mark.js
```diff
@@ -49,7 +48,7 @@ function Mark(props) {
 
   this._rule = new rules.VLSingle(type);
 
-  return Primitive.call(this);
+  // return Primitive.call(this);
```
deathbearbrown, line 51: should this just get removed now?

kadamwhite, line 51: Yep! I'll add it to my cleanup list.

deathbearbrown:
This is beautiful.  +1 functional test.  +1 code review.  I'm going to make sure we have an issue for the following known things:
- RECT - spacial properties
- ContentEditable react component

PR #311, opened 2016-03-25
set up scaffold, add placeholders. Get rid of inline styles

deathbearbrown:
Yay pretty!

I could iterate on this forever, so I'm going to stop.

@arvind check this out and play around with it.  Please let me know if you find any weirdness you really hate.  All I did was move things around and do some css clean up/ refactoring. But no major UX refactors in properties or pipelines. 

coolstuff

PR #337, opened 2016-04-10
Some UX Polish

arvind:

To better familiarize myself with the redux codebase, I started to apply some polish to the UX changes @deathbearbrown introduced with #311.
### Before

image
### After

image
### Changes proposed in this pull request
- Replaced Avenir with Lato. 
- Simplified the scaffolding grid. Previously, the widths of the sidebars were fixed, but padding was calculated via %s. Standardized the grid math to use pixels throughout. 
- Styled headers and borders in the inspectors to better demarcate groups of properties.
- Removed the `.indicator` element to give us more breathing space horizontally. When users hover over a data binding (scale/field), a strikethrough appears. Users can then click the binding to remove it, and return to a value/signal.
- Moved the `name` member of several Primitives to a private `_name` and a public `name()` method. This allows us to use just one field for the name, while ensuring that we escape/delimit spaces when exporting. This might have knock-on effects to @kadamwhite's PR #332, apologies! 
- Cleaned up the LayerList and GroupSubMenu components, to remove duplicated code.
- Inspector reducer also expands layers when a mark is selected. Reduces (;)) the need to explicitly repeat this functionality across various components. 
- Fixed bugs with selecting, expanding, renaming visuals and pipelines. Users click a pipeline name to expand (not toggle) it, double click to rename it. Users click a group/mark name to select it, double click to rename it, and click the spinner to expand/collapse groups.
- Reintroduced lost styling for scales.
- Replaced all FontAwesome `i` tags with SVG assets. Also introduced icons from the Noun Project. TODO: when we have an "about" modal dialog, we need to credit the Noun Project authors (details are currently commented within each of the assets).


arvind:
Seems like Karma is unhappy with how I slurp in the SVG assets. I'm using the [svg-sprite-loader](https://github.com/kisenka/svg-sprite-loader) webpack module. Perhaps there's a better way?

arvind:
Aha, got it! I was supposed to run the SVG assets through Karma's `raw-loader`.

deathbearbrown:
<3 
Ok the little feedback I have is:
1)  I think that for the groups we should keep the folders because groups are more like containers than menus, mentally.

2) The headers need to get brought out in that first column some how, or scene needs to be changed to look more like an action item.

3) the Groups +, Pipelines +  are not entirely centered in the column, They kind of break the over all grid of the design by being off like that.  It would look better if either they were styled in a completely different way, like a button that spans the whole width, or left aligned. 

4) the groups + and Pipelines + should be at the top of the column to avoid getting pushed off screen.
screen shot 2016-04-11 at 7 49 32 am

5) Clear all -  I would keep it in both places. The toolbar and the side bar, and then for the sidebar we need to give it a more "actionable" feel

6) scales needs to be absolutely positioned and z-indexed above groups, so that groups scrolls and doesn't knock scales out of the UI. 

The icons are all beautiful.  What you did the the property inspector is beautiful. 

I am adding a modal to my branch for walkthroughs, I can pull that out into it's own PR so you can work with it.

src/js/index.js
```diff
@@ -35,7 +35,7 @@ store.subscribe(throttle(model.update, 16));
 var ui = require('./components');
 
 // name model Scene
-model.Scene.name = 'Scene';
+model.Scene._name = 'Scene';
```
deathbearbrown: if we have a .name() to get the name we should have a setName() too,  also change name() to getName() ?

kadamwhite: I'm +1 to explicit getters and setters... this though I worry about b/c the way we construct a scene is a little different in my PR, why do we not just use .name like with the other primitives?

src/scss/inspector.scss
```diff
@@ -1,42 +1,48 @@
...
+$long-label-width: $label-width + 10px;
 
-  & > .indicator {
-    width: $can-drop-size;
-    height: $can-drop-size;
-    border-radius: 10px;
-  }
-
-  &.can-drop > .indicator {
+#inspector {
```
deathbearbrown, line 4: A css best practice is to never use IDs and never use html elements (h2, h3, dl, dt, button) and use classes instead. That is to prevent cases where you might the look of an element on a different element.  
So I might want the look of an h2, on my span.

src/js/index.js
deathbearbrown: because in this PR he changed .name to ._name with a getter method to retrieve it, unless I am misunderstanding your question.

kadamwhite: My question is, why was that change made. We have a considerable amount of code which assumes that the mark primitives (of which scene is one, even though it is handled a little differently) behave the same way in certain fashions, one of which is that they all can consistently be relied upon to have an `_id`, `type` and `name`. If we're changing to `_name` all over, I have a lot of work I need to do; and if we're doing it as a one-off for scene I want to understand why, because that has implications for the existing reducer code I've written. The motivation of the change is what I'm looking for.

arvind: Apologies for including this here. I realized it might cause knock on effects only after I pushed it. 

The reason I moved `name` to `_name` and exposed a method instead is because users can rename marks within the interface (double click on the name to make it a contenteditable). Users are likely to include spaces in names, but this is illegal in the Vega specification. So the `_name` member holds the exact value a user specified, and the function returns it with spaces replaced with a delimiter while exporting. 

If this is problematic, we can also introduce a `displayName` which is what Lyra 1 used, but I was hoping to avoid the unnecessary additional variable, 

As to why we're treating Scene differently? I'm not exactly sure. This particular line was already in the `lyra2` branch, and this commit just updated it to work with the renamed field. I can expose explicit separate getter and setter.

arvind:
Ok, just rebased this monster. A thorough functional test would be appreciated before merging @deathbearbrown, just to make sure I didn't screw something up. Thanks!

PR #339, opened 2016-04-11
add footer, modal, and move some containers around

deathbearbrown:
#337 has a todo to add a modal.  I have a modal and a place for about stuff in my walkthrough branch.  I pulled that out into this branch so we aren't doubling work.

- adds react modal
- adds footer
- moves a little bit of the markup around
- fixes an issue with the group sidebar not scrolling. 

- click on About in the footer. A placeholder modal should appear. 

The look of the footer is very basic and unstyled. 

modal

### DIFF
--- a/src/scss/modal.scss
+++ b/src/scss/modal.scss
@@ -1,0 +1,17 @@
+.ReactModal__Overlay{
+  top: '50%';
+  left: '50%';
+  right: 'auto';
+  bottom: 'auto';
+  transform: 'translate(-50%, -50%)';
+  z-index: 9999;
+  .closeModal {
+    position: absolute;
+    top: 20px;
+    right: 20px;
+    cursor: pointer;
+  }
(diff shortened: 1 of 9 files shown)


src/scss/modal.scss
```diff
@@ -0,0 +1,18 @@
+.ReactModal__Overlay{
+  top: '50%';
+  left: '50%';
+  right: 'auto';
+  bottom: 'auto';
+  marginRight: '-50%';
```
kadamwhite: This looks like a sass-ism, do we have any conventions in place around using this vs `margin-right`?

Do these values all need to be quoted?

deathbearbrown: haha I have that fixed in my walkthrough branch. I will fix here

deathbearbrown:
closing this since it's inside the walkthrough PR

PR #340, opened 2016-04-12
close tooltip on delete

deathbearbrown:
When a mark was deleted from the visualization sidebar, the tool tip wouldn't disapear. 

- remove the tooltip on delete
- also remove the tooltip on unmount of the component (to be double safe) 

1. make a group
2. delete it
3. tooltip should disappear

---
1. make a group
2. put a mark in the group
3. delete the mark in the group
4. tooltip should disappear

### DIFF
--- a/src/js/components/visualization/GroupSubMenu.jsx
+++ b/src/js/components/visualization/GroupSubMenu.jsx
@@ -1,6 +1,7 @@
 'use strict';
 var React = require('react'),
     connect = require('react-redux').connect,
+    ReactTooltip = require('react-tooltip'),
     ContentEditable = require('../ContentEditable'),
     lookup = require('../../model').lookup,
     hierarchy = require('../../util/hierarchy'),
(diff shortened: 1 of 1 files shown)


src/js/components/visualization/GroupSubMenu.jsx
```diff
@@ -72,12 +73,16 @@ var Group = connect(
...
+    ReactTooltip.hide();
     this.deleteMark(id);
     // set selected to null
     this.props.select(null);
     // redraw sidebar
     this.updateSidebar();
   },
+  componentWillUnmount: function(){
+    ReactTooltip.hide();
+  },
```
kadamwhite, line 85: If this is the failsafe, would it be sufficient to only call hide here, rather than in deleteUpdate?

deathbearbrown, line 85: No because the inner marks are created through a loop that makes the HTML. So when the loop is modified it doesn't actually unmount this component.

PR #341, opened 2016-04-12
Resolve pulse issue: Flow signal values to Lyra only through Vega

kadamwhite:
Up to now I had been pursuing a strategy by which all changes to signal values flowed into the Lyra store and the Vega view simultaneously. Given that the store listener called model.update but was throttled, and that setting values on the store from within the SignalValue mixin both dispatched singals to store AND (once model.update was fired) triggered callbacks that themselves dispatched that same signal to the store, the state of signals inside the store could get ambiguous and therefore the model itself could be put into an invalid state by rapidly triggering many signal changes one after another (e.g. scrubbing the color of a mark in a color picker).

This adds a "dispatch" parameter to the `sg.set` method, which can be provided to opt-out a call to `sg.set` from passing the value to the store as well. Having this explicit opt-out claries the flow of information within the SignalValue mixin by which we keep both Vega _and_ the Redux store up-to-date with signal values:

**Case 1: User interacts w/ mark through handles**
1. User drags a handle
2. Vega updates the signal value internally: _Vega is now up to date_
3. Vega fires an onChange handler for that signal
4. Lyra hears that onChange event and dispatches the new signal value to the Redux store: _Redux is now up to date_

**Case 2: User interacts w/ mark through inspector**
1. User sets a value in the inspector
2. Lyra sets the singal value within Vega, but _not_ Redux
3. Lyra calls `model.update()`, triggering a Vega update that makes the new value take effect: _Vega is now up to date_
4. Vega fires an onChange handler for that signal
5. Lyra hears that onChange event and dispatches the new signal value to the Redux store: _Redux is now up to date_

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,6 @@
     "karma-sourcemap-loader": "^0.3.6",
     "karma-spec-reporter": "0.0.23",
     "karma-webpack": "^1.7.0",
-    "lodash.debounce": "^4.0.3",
-    "lodash.throttle": "^4.0.1",
     "minami": "^1.1.1",
     "mocha": "^2.4.5",
     "node-sass": "^3.4.2",
--- a/src/js/components/inspectors/Property.jsx
+++ b/src/js/components/inspectors/Property.jsx
@@ -5,25 +5,19 @@ var React = require('react'),
     ContentEditable = require('../ContentEditable'),
     model = require('../../model'),
     lookup = model.lookup,
-    resetProperty = require('../../actions/ruleActions').resetProperty,
-    vegaInvalidate = require('../../actions/vegaInvalidate'),
-    addVegaReparseRequest = require('../mixins/addVegaReparseRequest');
+    resetProperty = require('../../actions/ruleActions').resetProperty;
 
 function mapDispatchToProps(dispatch) {
   return {
     resetProperty: function(id, property) {
       dispatch(resetProperty(id, property));
-    },
-    requestReparse: function() {
-      dispatch(vegaInvalidate(true));
     }
   };
 }
 
 var Property = React.createClass({
   propTypes: {
-    resetProperty: React.PropTypes.func,
-    requestReparse: React.PropTypes.func
+    resetProperty: React.PropTypes.func
   },
 
   mixins: [SignalValue],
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -62,8 +62,6 @@
     "karma-sourcemap-loader": "^0.3.6",
     "karma-spec-reporter": "0.0.23",
     "karma-webpack": "^1.7.0",
-    "lodash.debounce": "^4.0.3",
-    "lodash.throttle": "^4.0.1",
```
kadamwhite, line 66: I removed the throttling on the store listener, since it now doesn't call model.update except when the selected mark changes there is very little logic that would run each digest so the impact is minimal. These tools got us through earlier hurdles but the value of these methods has finally outweighed their cost by obscuring the model state sync issue.

src/js/components/inspectors/Property.jsx
```diff
@@ -5,34 +5,26 @@ var React = require('react'),
     ContentEditable = require('../ContentEditable'),
     model = require('../../model'),
     lookup = model.lookup,
-    resetProperty = require('../../actions/ruleActions').resetProperty,
-    vegaInvalidate = require('../../actions/vegaInvalidate'),
-    addVegaReparseRequest = require('../mixins/addVegaReparseRequest');
```
kadamwhite, line 10: This is only tangentially related to this PR; individual actions that would invalidate Vega (adding or removing marks, binding data, etc) handle the invalidation themselves, so the explicit invalidator call is not needed here.

src/js/components/mixins/SignalValue.jsx
```diff
@@ -74,7 +74,9 @@ module.exports = {
     }
 
     if (signal) {
-      sg.set(signal, value);
+      // Set the signal on the view but do not dispatch: the `.signal` listener
+      // above will dispatch the action to synchronize Redux with Vega.
+      sg.set(signal, value, false);
```
kadamwhite, line 79: I went back and forth on whether to make this behavior default but I think that having this be "false" (and implicitly "true" if not set) leads to the clearest mental picture of the flow of operations within this file

kadamwhite:
Got +1 from @arvind in Slack

PR #343, opened 2016-04-12
Dispatch actions when extentproperty handles a change to fix bug

kadamwhite:

This fixes an issue introduced earlier where the ExtentProperty UI is broken.

- Use disable/enableProperty actions to toggle disabled state on rect extent properties

rect extent

@arvind I could use your testing to see whether what's happening is what we expect, esp. when you bind data!


src/js/components/inspectors/AreaProperty.jsx
```diff
@@ -53,33 +53,6 @@ var AreaProperty = React.createClass({
...
-
-    if (value === center && state.end !== span) {
-      update[state.end]._disabled = true;
-      update[span]._disabled = false;
-      state.end = span;
-    }
-
-    this.requestVegaReparse();
-  },
-
```
deathbearbrown, line 82: is this just clean up from the last branch?

kadamwhite, line 82: This looked like unused copy-paste from your initial commit adding the area inspector... if I'm wrong I'll reinstate it but that method wasn't being called as it is in extentproperty

deathbearbrown:
A bug, which I have recreated 3 times now, but unpredictably. 
screen shot 2016-04-13 at 7 42 48 am
- Add a rect.  
- mess around with the spacial properties.  
- Add data to each axis
- mess around with the spacial properties. 
- Add another mark.
- Delete that mark.
- Try to click on the rect in the sidebar.

This is what the properties look like when I get the bug:
screen shot 2016-04-13 at 7 48 14 am

You'll know things are broken because the graph looks weird.

@arvind have you ever seen this in vega?

src/js/components/inspectors/AreaProperty.jsx
deathbearbrown, line 82: yesssss i remember what I was gonna do with this.  There's a bunch of conditional stuff with area marks I didn't get to finish.  

this is good cleanup

deathbearbrown:
+1 code.  I did functional testing and other than a really hard to recreate bug that I don't think is a blocker, I am +1 functional.

Waiting on Arvind to +1 functional too since he knows what the behavior should look like.

arvind:
This is functional close for me. Couple of bugs I'm noticing:
1. After adding a rect mark, if I switch any of the spatial properties (e.g., "right" --> "width") the two corresponding textboxes are empty (even though an actual value is being used to render the visualization). If I enter values into the textboxes, the visualization updates correctly.
   
   rectspatial

~~2. The `SpatialPreset` component for width/height isn't being rendered anymore? Just as we see a checkbox for "Group width" and "Group height" when using left/right or top/bottom respectively, we should see a checkbox saying "Automatic" when width/height are selected.~~

arvind:
Jk on (2) above. The "Automatic" only shows up when width/height is bound to a scale. However, (1) continues to be a problem. We also have a regression from the current `lyra2` branch. Say we bind a field to a rect's horizontal channel. In `lyra`, the inspector updates correctly to show that `Left` and `Width` have been set. In this branch, however, `Left` shows as set but `Right` is still selected (if we manually switch `Right` to `Width`, we see the correct scale/field binding appear, but this should happen automatically).

arvind:
Fixed with f593ab8.

PR #345, opened 2016-04-13
Store Listener cleanup

kadamwhite:

This PR does two unrelated things (sorry): First, it makes it so that when you add a mark, it auto-expands that mark and/or the mark's parent in the inspector sidebar:

Before:

mark-expand-before

After:

mark-expand-after

This also makes a small but conceptually valuable change to the work done in `mapStateToProps`: wherever possible `mapStateToProps` should not be used for any purpose beyond **reading values from the store**, and **should only return values comparable by basic equality**.

To elaborate, `mapStateToProps` fires on every digest cycle; the component's Render will only fire when the values returned from `mapStateToProps` change, where "change" is determined by equality. For this reason, we should preference returning basic types from `mapStateToProps` (string, number, boolean) and Immutable constructs, rather than objects or arrays, because all of the former (including Immutable constructs) can be compared by equality: any sort of `.toJS`, `.toObject`, etc, which generates a value that cannot be compared by equality, should be handled in `.render` and not in `mapStateToProps`, to avoid equivalent objects like `{1: true}` and `{1: true}` being viewed as "different" values and therefore triggering unnecessary (and therefore wasteful) re-renders.

- Auto-expand newly-added marks and their parents
- Return basic types and Immutable constructs from `mapStateToProps`, not objects or arrays
- Only include `ownProps` in map\* methods when necessary

- The usual

### DIFF
--- a/src/js/components/InspectorSidebar.jsx
+++ b/src/js/components/InspectorSidebar.jsx
@@ -18,9 +18,7 @@ function mapStateToProps(reduxState, ownProps) {
   };
 }
 
-var Inspector = connect(
-  mapStateToProps
-)(React.createClass({
+var Inspector = React.createClass({
   classNames: 'sidebar col5 push4 md-blue-bg',
   render: function() {
     var props = this.props,
@@ -60,11 +58,12 @@ var Inspector = connect(
       </div>
     );
   }
-}));
+});
 
-module.exports = Inspector;
 Inspector.Line = require('./inspectors/Line');
 Inspector.Rect = require('./inspectors/Rect');
 Inspector.Symbol = require('./inspectors/Symbol');
 Inspector.Text = require('./inspectors/Text');
 Inspector.Area = require('./inspectors/Area');
+
+module.exports = connect(mapStateToProps)(Inspector);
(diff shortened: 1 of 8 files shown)


deathbearbrown:
I had to update my notes for this.  But +1 code.  I fixed my walkthrough branch so I'm not doing any .toJS() stuff

deathbearbrown:
+1 functional.  Waiting for Arvind to check it out before merging.

src/js/components/InspectorSidebar.jsx
```diff
@@ -60,11 +58,12 @@ var Inspector = connect(
...
+});
 
-module.exports = Inspector;
 Inspector.Line = require('./inspectors/Line');
 Inspector.Rect = require('./inspectors/Rect');
 Inspector.Symbol = require('./inspectors/Symbol');
 Inspector.Text = require('./inspectors/Text');
 Inspector.Area = require('./inspectors/Area');
+
+module.exports = connect(mapStateToProps)(Inspector);
```
arvind, line 69: I like this new style better!

arvind:
👍 Mergin'

Issue #342, opened 2016-04-12
BUG: Content editable react component


deathbearbrown:
- [x] I am reporting a bug

**Expected Behavior**
when I click on the name of a mark in the visualization sidebar, I can update the name, but there is the following error in the console:

```
warning.js:45 Warning: A component is `contentEditable` and contains `children` managed by React. It is now your responsibility to guarantee that none of those nodes are unexpectedly modified or duplicated. This is probably not intentional.
```

kadamwhite:
Reading related to the error itself: https://github.com/facebook/react/issues/2477 (and by extension https://medium.com/medium-eng/why-contenteditable-is-terrible-122d8a40e480#.aq2xacp42 ) -- we've made the choices we did for a reason, but contenteditable is basically always going to shout at us in the console. The main purpose of this ticket then becomes, "make it so that the change persists throughout the app."

PR #346, opened 2016-04-13
De-couple contenteditable from signalvalue and use store to save names

kadamwhite:

The ContentEditable mixins had not been updated to persist the updated strings in the main lyra store, meaning that mark name changes were not consistently represented throughout the application.

This also cleans up the way SignalValueMixin had been overloaded to handle ContentEditable data persistence: Instead, `<ContentEditable>` now takes a `save={someFn}` attribute, and that save method is called with the updated value of the editable area.

In cases where this value is a mark name that is represtented in the store, we use `save` to dispatch an action that sets the value in the store; and for now we also set the value on the primitive itself, since primitives are only sync'd from the store when a change occurs that would warrant a Vega re-parse. (And for now I believe renaming a mark is not in that category).

If the value being updated is _not_ yet part of the store, we simply define a contextual `save` method that will set the value on whatever instantiated primitive we want to modify.

- Add an `updateMarkProperty` action dispatcher & reducer
- De-couple ContentEditable from SignalValueMixin
- Update ContentEditable to use a save method passed in via property attribute
- Set ContentEditable components to update their values in instances or the store or both

**This PR includes:**
- [x] Tests
- [x] functional comments
- ~ high level description of any new components

- Create a mark; rename it; observe that the name is updated in the inspector sidebar
- Create a group; create a mark within it; rename the mark; close its parent group; reopen the group; observe that the mark's name is still what you entered
- Rename a pipeline; use the `window.primitives` global in the console to observe that the name is saved to the object
- Bind data; rename a scale; use the `window.primitives` global in the console to observe that the name is saved to the object


kadamwhite:
For #342

src/js/components/visualization/ScaleList.jsx
```diff
@@ -29,12 +29,15 @@ var ScaleList = React.createClass({
         </h4>
         <ul>
           {props.scales.map(function(scale) {
+            function updateScaleName(val) {
+              scale.name = val;
+            }
```
kadamwhite, line 34: This approach will make it easy for us to replace these with action dispatcher component properties once scales are in the store

deathbearbrown:
ok + 1 code.  Functional tested ~~BUT... question.~~

~~Do we need to validate characters/escape any characters that can be in names?~~
it ain't breaking

kadamwhite:
@deathbearbrown thanks! To answer your question, yes, but after speaking with @arvind in Slack he's got an approach for handling that which I can live with. My overall instinct is to eventually have the reducer set both a `_name` (verbatim, used in Lyra UI) and `name` (escaped/transformed, used in Vega output) properties on each mark, scale, etc. Doing that now was infeasible because guides and such are not in the purview of Redux.

src/js/components/inspectors/Property.jsx
```diff
@@ -106,7 +103,6 @@ var Property = React.createClass({
             <div>
               <input type="color" value={!disabled && value} disabled={disabled}
                 onChange={this.handleChange} />
-              <ContentEditable {...props} />
```
arvind, line 109: Just got caught these. Is there a reason they were removed? Users quite liked being able to manually enter hexcodes/numbers rather than being forced to use a color picker/slider.

kadamwhite, line 109: My bad, I couldn't see them in the UI or didn't know how to use them, or because I was gravitating towards the color picker... since it didn't visibly change the UI to remove them and it made my ContentEditable vs SignalValue refactor easier to remove them, I did so. The use-case you describe makes sense and I'd be in favor of reinstating them.

PR #347, opened 2016-04-13
Remove deprecated .parent and .child prototype methods

kadamwhite:

`Group.prototype.child` was deprecated last week; now that Redux is meant to be the source of truth for marks, the proper way to manage hierarchy is to dispatch a `setParent` action or take some related action that lets the reducer handle setting the `.marks` and `._parent` properties used to manage the hierarchy of Lyra primitives.

If `.child()` is no more, then related utilities like `getChildren` (which was written provisionally, then never used) can also be removed; and at that point it seemed best to remove `Primitive.prototype.parent` as well, in favor of always using the `getParent` utility method.

In areas where `.parent()` was being used to set parent on assignment (via the construct `result = new Constructor(args).parent(parentId)`), I have added `_parent` as an optional propery to the relevant constructor so that the value can be set in one statement without needing to chain any method calls together.

- Remove `Group.prototype.child`
- Re-write hierarchy test setup code to avoid needing `.child` method
- Use `getParent` utility in place of `.parent()` prototype method
- Remove `Primitive.prototype.parent`
- Augment constructors still outside of Redux's purview to accept a `_parent` ID value when necessary
- Remove unused `hierarchy.getChildren` utility

**This PR includes:**
- [x] Tests
- [x] functional comments

- Functional test: Did anything break?

### DIFF
--- a/src/js/model/primitives/Primitive.js
+++ b/src/js/model/primitives/Primitive.js
@@ -4,7 +4,6 @@
 var dl = require('datalib'),
     sg = require('../signals'),
     model = require('../'),
-    lookup = model.lookup,
     counter = require('../../util/counter');
 
 /**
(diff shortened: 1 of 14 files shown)


deathbearbrown:
+1 code and +1 functional test, nuffin borked. is all good

PR #348, opened 2016-04-13
Use "Mark" in place of "Primitive" in redux store, actions, etc

kadamwhite:
Per review, we should err on the side of specificity in naming when possible.

**This could strongly use a very close reading,** to make sure I didn't rename something I shouldn't have!


src/js/store/listeners.js
```diff
@@ -27,7 +27,7 @@ var sg = require('../model/signals'),
     parseInProgress = require('../actions/vegaParse');
 
 function instantiatePrimitivesFromStore(store, model) {
-  store.getState().get('primitives').forEach(function(props, markId) {
+  getIn(store.getState(), 'marks').forEach(function(props, markId) {
```
deathbearbrown, line 30: Hey @arvind , when future people join the project who are new to redux & immutable.js, make sure you explain the what the immutable utility is.  That way when they look at the code and see the getIn()  for one level fetches from the store eg. `store.getState().get('mapname')` they don't get confused about it not matching the immutable docs.

deathbearbrown:
Fine tooth comb on this and I can't find any regressions concerning this.   I found 2 regressions for other things that I'll make PRs for in a second :+1:

PR #350, opened 2016-04-14
scene spec not returning height & width

deathbearbrown:
I noticed this happening in the UI when I added a new group to the scene and decided to investigate:
wat

Turns out that the way the signals were added to the scene in the reducer was breaking setting the width/height.

I removed that and also updated the scene action and reducer (and tests) to let you initialize with new properties.  It currently wouldn't let you do that.

I also added a test to make sure that height and width always returned for scene when exporting.

wat2
You can see a tiny jump because of the padding in the scene, but it's not as drastic as it was before.

**This PR includes:**
- [x] Tests

### DIFF
--- a/src/js/model/primitives/marks/Scene.js
+++ b/src/js/model/primitives/marks/Scene.js
@@ -1,11 +1,9 @@
 'use strict';
 var dl = require('datalib'),
     inherits = require('inherits'),
-    sg = require('../../signals'),
-    Group = require('./Group');
-
-var SG_WIDTH = 'vis_width',
-    SG_HEIGHT = 'vis_height';
+    Group = require('./Group'),
+    store = require('../../../store'),
+    getIn = require('../../../util/immutable-utils').getIn;
 
 /**
  * @classdesc A Lyra Scene Primitive.
@@ -38,6 +36,7 @@ Scene.prototype.parent = null;
  * @returns {Object} The default mark properties
  */
 Scene.defaultProperties = function(props) {
+  console.log(props);
   // Note that scene has no "properties" property
   return dl.extend({
     // Containers for child marks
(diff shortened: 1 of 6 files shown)


src/js/model/primitives/marks/Scene.js
```diff
@@ -38,6 +38,7 @@ Scene.prototype.parent = null;
  * @returns {Object} The default mark properties
  */
 Scene.defaultProperties = function(props) {
+  console.log(props);
```
kadamwhite, line 39: Debugging code

src/js/reducers/signals.test.js
```diff
@@ -249,6 +249,22 @@ describe('signals reducer', function() {
           _idx: 1
         }
       });
+    });
+
+    it('initializes the scene width & height signals set to custom values', function() {
+      var result = signalsReducer(initialState, createScene({width:400, height:100}));
```
kadamwhite: nit but this'd be more readable if the width and height were indented out onto their own lines

kadamwhite:
The reducers were implemented as they were because the scene's init method has always initialized and bound to the vis_height and vis_width signals; if I'm reading this correctly, we're no longer associating the scene with the vis height and width signals, so it's not clear how to change the scene's dimensions as a whole. I'd be curious for more information on the specific issue in the reducer that was causing the large jumps, out of curiosity.

deathbearbrown:
the vis_height and vis_width signals never existed on the vega view when you would initialize the scene though doing it like that. 

 In the OLD OLD OLD code the vega signals were initialized in Scene.js before doing property stuff.

deathbearbrown:
oh wait I KNOW WHAT'S UP.

kadamwhite:
> the vis_height and vis_width signals never existed on the vega view when you would initialize the scene though doing it like that.

Just discussed in slack w/ @deathbearbrown, and confirmed her suspicion: the bug originated because I altered the export code to run AFTER we destroy the scene, but the Scene's .export method tries to use `sg.get` to read the width/height off of vega. Switching those accessors to point at the store instead should resolve the issue.

deathbearbrown:
Ok this is all linted and updated and stuff. I removed sg from Scene.js and am using the store. :)  Added the scene parts back to the reducer.

I have a better idea about what still needs doing. Thank you KAdam for letting me ramble at you.

arvind:
Merged via the command line as 760dd5f.

PR #351, opened 2016-04-14
remove Marks.init

deathbearbrown:
This is a cleanup PR.
We no longer use the init function in Marks.js since it's all handled in the store. (I double checked with KAdam)  I also removed Marks.test.js which had skipped tests in it for stuff that no longer exists. WOOO. Deleting.


kadamwhite:
:+1:

PR #355, opened 2016-04-18
imperfect fix for drag and drop in firefox

deathbearbrown:
I don't want to spend a lot of time troubleshooting every part of this, so here's what I got.  Drag and drop works now.  The ghost element does not look right.

Soo fun.


arvind:
Addressed in #337.

PR #356, opened 2016-04-18
deleteUpdate is not being used, breaks in UI branch

deathbearbrown:
This fixes an issue with delete marks in the sidebar.  I think one of our many rebases might have messed this up.

The issue was you could not delete a mark inside a group. Also deleting a group or mark would cause a lot of console errors.

Also removed a nested ternary operator I should have caught in the UI branch. SHAME ON ALL OF US. 

To test this, delete a lot of things.

### DIFF
--- a/src/js/components/InspectorSidebar.jsx
+++ b/src/js/components/InspectorSidebar.jsx
@@ -29,9 +29,10 @@ var Inspector = React.createClass({
   render: function() {
     var props = this.props,
         // props.id existence check handles the initial application render
-        primitive = props.id ? lookup(props.id) : {},
-        from = lookup(primitive.from),
-        ctor = primitive.constructor.name,
+        primitive = props.id ? lookup(props.id) : {};
+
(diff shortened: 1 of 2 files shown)


src/js/components/InspectorSidebar.jsx
```diff
@@ -29,9 +29,10 @@ var Inspector = React.createClass({
   render: function() {
     var props = this.props,
         // props.id existence check handles the initial application render
-        primitive = props.id ? lookup(props.id) : {},
-        from = lookup(primitive.from),
-        ctor = primitive.constructor.name,
+        primitive = props.id ? lookup(props.id) : {};
+
+    var from = primitive ? lookup(primitive.from) : '',
+        ctor = primitive ? primitive.constructor.name : '',
```
kadamwhite, line 35: Might be good to add a comment indicating that these checks are only relevant if props.id is not found in the primitive store, b/c once the store evolves further the conditional logic here may be obviated

Issue #362, opened 2016-04-28
serving on another port and not defaulting to "localhost"


aliabbasjp:
How can I serve lyra so that it can be accessed remotely currently remote access is not possible : 

both 
http://localhost:8080/

http://<ip of server>:8080/

refuses connection remotely eventhough its on the VPN and it works fine on the server itself

PR #363, opened 2016-04-28
Deployment

kadamwhite:

This adds a `npm run build:prod` script, which uses a new production-mode wrapper for the webpack config to set certain optimization options useful when deploying Lyra as a standalone app on a static file server.

Partially addresses #362 by providing tools for generating a truly standalone build: webpack-dev-server was never intended for production use.

The key changes in a production-mode build:
1. Use of the [`-p` Webpack config option](https://webpack.github.io/docs/cli.html#production-shortcut-p) to enable Webpack's build optimizations
2. Use of a different sourcemap option
3. Use of a DefinePlugin to globally hard-set a "production mode" constant to permit Redux and other code to opt-in to performance-tuned builds instead of development-tool builds
4. CSS content is rendered to a separate output file; In order for Lyra to be deployed on a web server, it is most efficient to have each of the code bundles separated out into scripts or styles as appropriate.

With this change, a complete "deployment" workflow will look like this:
1. `npm run build:prod` to trigger a build in production mode
2. Copy `index.html`, `build/`, and `data/` to a server (or e.g. to the Github Pages branch)
3. Lyra should now be usable on that server

Note that this commit removes the plugin that was being used to copy the index HTML into the build directory, as that step was superfluous; the index.html from the repository root is the one with the correct paths for use with built assets.

@arvind Testing this locally works just as well as the lyra2 branch currently does, but pipelines are showing up in neither branch right now so it's hard to thoroughly functionally test.


deathbearbrown:
Heyas @kadamwhite I tested on Lyra2 and this branch and the pipelines are showing up for me. I am going to do a clean checkout of the project. NPM is just installing mad slow today.

arvind:
Shoot, I just merged this into master but I noticed that inspectors for marks aren't showing up on a production build anymore. Pipelines and the rest of the interface seems to be fine though.

kadamwhite:
Oh hmm interesting. I'll take a look

PR #365, opened 2016-05-24
Move pipelines to store [WIP]

nikhilkhanna:
This pull requests moves the pipelines (not the datasets yet) to the store and wires up the react components to function with these new pipelines.

To test this, you can simply attempt to bind data to a mark and notice that the pipeline objects are now objects in the redux store, but overall functionality is still maintained.


src/js/components/PipelinesSidebar.jsx
```diff
@@ -1,13 +1,21 @@
...
+    getIn = require('../util/immutable-utils').getIn,
     PipelineList = require('./pipelines/PipelineList'),
     assets = require('../util/assets'),
     Icon = require('./Icon'),
     model = require('../model');
 
+function mapStateToProps(reduxState) {
+  var pipelines = getIn(reduxState, 'pipelines.pipelines');
+  return {
+    pipelines: pipelines.toJS(),
```
arvind: See kadam's comment in [this PR](https://github.com/vega/lyra/pull/345) about using `toJS` in `mapStateToProps`. Essentially, `toJS` can be an expensive function, so we want to use it as infrequently as possible. `mapStateToProps` should return something quick that can be reliably used in a straight equality test to determine whether to re-render the property.

src/js/reducers/pipelines.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {Object} state - An Immutable.Map state object
+ * @param {Object} action - A redux action object
+ * @returns {Object} A new Immutable.Map with the changes specified by the action
+ */
+function pipelinesReducer(state, action) {
+  if (typeof state === 'undefined') {
+    return Immutable.fromJS({
+      pipelines: []
+    });
+  }
```
arvind, line 30: Because we use [combineReducers](http://redux.js.org/docs/api/combineReducers.html) in `reducers/index.js`, the `state` that is received by this reducer only contains our pipelines state. So we don't need to further nest it in here. 

Also, rather than using an array to store the pipelines, lets switch to an `ImmutableMap`. The keys in this map will be the pipeline IDs, and saves us iterating through the array to find the pipeline we care about for other actions (e.g., your L50 codeblock). Take a look at the `marks` or `signals` reducers for an example.

src/js/reducers/pipelines.js
```diff
@@ -0,0 +1,74 @@
...
+
+  var pipelines = state.get("pipelines")
+
+  if (action.type === actions.CREATE_PIPELINE) {
+    var newPipelines = pipelines.push(Immutable.fromJS({
+      name: action.id,
+      id: counter.type("pipeline"),
+      source: null
+    }));
```
arvind, line 40: This looks good to me for the most part. One thing to keep in mind with objects in our store: the keys/properties of the object should correspond to things found in the output Vega specification. Any Lyra-specific things should be prefixed with an underscore. The `export` methods then automatically strip these Lyra-specific properties out. 

It matters less with Pipelines since we never export them directly (they're just a helper structure to group together related Datasets). But, for consistency, we should use `_id` instead of `id`. `source` is probably fine as is. 

If in doubt, follow the naming conventions found in the existing class definitions or feel free to ping me via Slack :)

arvind:
Hi @nikhilkhanna, I've merged some older PRs in from the end of the Bocoup engagement and that's created some conflicts on this branch. Could you rebase and push so that I can merge w/o conflicts? It would also be good to get rid of the three unrelated commits at the head of this PR (regarding the Dockerfile). Thanks!

Issue #368, opened 2016-06-01
Bug with delete marks and add marks


deathbearbrown:
**Expected Behavior**
- add 2 marks
- delete one
- add new mark & the mark will show up in the visualization side bar, the canvas & be exported

**Actual Behavior**
- add 2 marks
- delete one
- add new mark, 
  - inspector is there
  - doesn't show up on the canvas
  - not in visualization sidebar 
  - not in the export

This looks like a syncing issue, I am looking into it.

deathbearbrown:
More bug updates:
- happens with primitives applied directly to the scene.  Primitives in a group can be deleted and re-added and it's NBD

deathbearbrown:
@arvind good news, it is not a sync issue. The parent was just being passed as null from the toolbar react component :/ boo. That's why groups always worked because it was set up in a different react component. 
PR here #371

arvind:
Fixed with #371, thanks @deathbearbrown!

PR #371, opened 2016-06-01
fixes 368 issue with parent being unset after delete

deathbearbrown:
Fixes issue #368

- add a check to see if the parent being passed is null and select sceneid
- bind `this` to the toolbar's addMark function

- Add 2 marks to the scene 
- delete one of the marks
- add a new mark to the scene
  - you should see the mark show up on the canvas and in the visualization's sidebar

### DIFF
--- a/src/js/components/toolbar/AddMarks.jsx
+++ b/src/js/components/toolbar/AddMarks.jsx
@@ -45,9 +45,7 @@ var AddMarksTool = React.createClass({
   render: function() {
     // Closest container is determined by walking up from the selected mark,
     // otherwise it defaults to the scene itself
-    var closestContainerId = this.props.selectedId ?
-      getClosestGroupId(store.getState(), this.props.selectedId) :
-      this.props.sceneId;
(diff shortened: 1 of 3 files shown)


src/js/components/toolbar/AddMarks.jsx
```diff
@@ -54,7 +55,7 @@ var AddMarksTool = React.createClass({
         {marksArray.map(function(markType, i) {
           return (
             <li key={markType}
-              onClick={this.props.addMark.bind(null, markType, closestContainerId)}>
+              onClick={this.props.addMark.bind(this, markType, closestContainerId)}>
```
arvind: Hmm, but I'm curious why `closestContainerId` isn't set the right thing? From the looks of it, we don't clear out `inspectors.selected` in the redux store on `MARK_DELETE`. As a result, `closestContainerId` isn't set to the `sceneId`. Would that be a better/cleaner fix?

deathbearbrown:
updated to return scene id instead of null :)

kadamwhite:
:+1: this is an even better fix than what I'd experimented with!

Issue #375, opened 2016-06-07
Scales with Rect Mark 


deathbearbrown:
There are a couple bugs around scales and Rect Marks. 

The first is a vega error.  You can recreate this by doing the following:
![screen shot 2016-06-07 at 8 00 28 am](https://cloud.githubusercontent.com/assets/607541/15871697/d023a50c-2caa-11e6-9a15-42e44b1ecc1a.png)
- add a rect mark
- attach data to each axis twice
- try to attach data to one axis for a third time

This happens sometimes on the 2nd time you try to attach data to an axis, but I can reliably recreate this bug by doing it to an axis 3 times. 

The second bug is about reusing scales that were originally set for a rect.  It's not so much a bug as a problem that needs to be worked out. Steps to see the problem:
- add a rect mark
- attach POPULATION to Y axis
- delete rect mark
- add a line
- attach POPULATION to Y axis
  Now you have an Y, and a Y2 scale

If you do that with 2 linear marks like Symbols and Lines, you will just have the one Y scale.

PR #376, opened 2016-06-07
Move Scales to redux

deathbearbrown:
This is part one of the refactor.  It took a while because I kept finding bugs with scales in general. 
Some of them are reflected in issue #375 

And additional bug/problem is how we want to handle scales when we've already added x1, x2, y1, y2.  Can you have more scales than that??

To test this, add data to a mark (test for multiple axi) and you should see scales show up in the sidebar.

I haven't added tests yet because I want to refactor this more.

Next I am going to add:
- clearing out/ removing scales
- a simple inspector


arvind:
> And additional bug/problem is how we want to handle scales when we've already added x1, x2, y1, y2. > Can you have more scales than that??

Yes, you can have an unlimited number of scales (x, y, etc.). E.g., you may have multiple different marks drawing from different datasets as part of the same visualization.

src/js/components/visualization/ScaleList.jsx
```diff
@@ -7,24 +7,30 @@ var React = require('react'),
...
 function mapStateToProps(reduxState, ownProps) {
   return {
-    selectedId: reduxState.get('selectedMark')
+    selectedId: reduxState.get('selectedMark'),
+    scales: reduxState.get('scales')
   };
 }
 
+function mapDispatchToProps(dispatch) {
+  return {};
```
arvind, line 16: If this is empty, do we still need it?

arvind:
One functional bug I've noticed: if I drag out a rect mark and then some ordinal data (e.g., gapminder > country) to x for example, the inspector updates to show the scale/field for "left", but "right" doesn't update to "width" and automatic. On lyra2, this works correctly after accounting for the "forceUpdate" (e.g., after closing out the hints box). This might be related to #375 or even #343?

Except for this, 👍 for functionality.

untitled

src/js/components/visualization/ScaleList.jsx
deathbearbrown, line 16: no, it's just a place holder for when you change the scale name.

PR #377, opened 2016-06-08
WIP: use redux primitives for extents

deathbearbrown:
Problem:

> One functional bug I've noticed: if I drag out a rect mark and then some ordinal data (e.g., gapminder > country) to x for example, the inspector updates to show the scale/field for "left", but "right" doesn't update to "width" and automatic.
> 

@arvind  can you please test branch and just check that I actually fixed this bug?  I think something might still be wonky.


arvind:
Fixed with f593ab8.

PR #378, opened 2016-06-10
Scales inspector

deathbearbrown:
- Re styles the scale sidebar
- shows the inspector on clicking
- unselects any active marks
- passes scale data to the inspector (but form is not there yet, just needs someone to set that up)

**This PR includes:**
- [x] Tests
- [x] functional comments

- add a mark
- attach data to it
- scale should show up in sidebar
- click on scale
- inspector placeholder should show up for scale with information about the scale
- click on mark - mark inspector should show


src/js/actions/addScale.js
```diff
@@ -9,7 +9,7 @@ function addScale(data) {
   }, data);
 
   var action = {
-    id: props._id,
+    id: 'scale_' + props._id,
```
arvind, line 12: @deathbearbrown: Why is the `scale_` prefix needed here?

src/js/reducers/inspector.js
```diff
@@ -21,7 +21,11 @@ function inspectorReducer(state, action) {
         selected: null
       },
       selected: null,
-      expandedLayers: {}
+      expandedLayers: {},
+      scales: {
+        selected: null,
+        show: false
+      }
```
arvind, line 28: @deathbearbrown Another question -- why model scale inspectors separately here? Since only a mark or scale can be selected at any time, can we not just reuse `selected` at the top-level? The pipelines sidebar has a different selection behaviour, so breaking that out separately makes sense. 

This is probably good enough for now, pending a refactor of how the inspectors state is modeled in the store.

src/js/actions/addScale.js
deathbearbrown, line 12: I was having issues retrieving from the store with a key as an integer. For marks all the ids are always strings even when they are numbers.  I added that to test this theory and make things work so I could get the PR in asap.

src/js/reducers/inspector.js
deathbearbrown, line 28: Because selected is just the ID for marks and I didn't want to break existing mark functionality, since there are listeners all over for that. 
Again, this is a little rushed. I figure there will be a clean up of redux once everything gets in there.

src/js/actions/addScale.js
kadamwhite, line 12: The `get` and `set` utilities I wrote coerce all keys to strings, and were intended to resolve that issue

arvind, line 12: Thank you all!

PR #381, opened 2016-06-16
Refactor model to redux

arvind:
These commits move the definitions for datasets into the redux store. A set of dataset utility methods load the raw values and compute the schema external to redux. It's marked as a WIP as production rules and export logic heavily relies on datasets being in the old model. So, this PR should not be merged until that code has been refactored as well.


arvind:
@kadamwhite, @deathbearbrown Say goodbye to the old `model`! 🎉

deathbearbrown:
WOOOOOOO!!!! 🎈🎈🎈🎈

kadamwhite:
woo! 🚀

Issue #331, opened 2016-04-06
QA testing flows


deathbearbrown:
When doing QA functional testing (of the mtuk variety) there are some flows of current functionality that should not be broken.  Add more flows as the app development progresses. 

Note:  It might seem like overkill to have 2 different flows for adding data to one axis and then data to 2 axis, but this will account for any bugs that might happen if only one scale exists. 
### Add a new mark

Do for each of the following marks:  LINE, RECT, AREA, TEXT, SYMBOL
#### Scene level

Scenario 1:   
- add a mark to the scene
- drop data on an axis
- delete

**Expected outcome:**  
- On add, the new mark will be selected.  
- When data is moved over the mark, the drop zones will appear.  
- It will be highlighted in the sidebar.
- When dropped the Scales sidebar will update with one of the following: X, X1, Y, or Y1. 
- On delete the mark will disappear, but the scales for x & y will remain on the canvas and in the sidebar

Scenario 2:
- add a mark to the scene
- drop data on each axis
- delete

**Expected outcome:**  
- On add, the new mark will be selected.  
- When data is moved over the mark, the drop zones will appear.  
- It will be highlighted in the sidebar.
- When dropped the Scales sidebar will update with two of the following: X, X1, Y, or Y1. 
- On delete the mark will disappear, but the scales for x & y will remain on the canvas and in the sidebar

Scenario 3:
- add a mark to the scene
- drop data on an axis
- modify a property on the mark (e.g. color)
- delete

**Expected outcome:**   
- On add, the new mark will be selected.  
- When data is moved over the mark, the drop zones will appear.  
- It will be highlighted in the sidebar.
- When dropped the Scales sidebar will update with X, X1, Y, or Y1. 
- Properties will update when modified
- On delete the mark will disappear, but the scales for x & y will remain on the canvas and in the sidebar
### Group level

Scenario 1:
- add a group 
- add a mark to the group
- drop data on the x axis
- delete mark

**Expected outcome:**  
- On add, the new group will be selected.  It will be highlighted in the sidebar. 
- On add, the new mark will be selected.  It will be highlighted in the sidebar.  
- Mark will nest under the group tab in the menu. 
- When data is moved over the mark, the drop zones will appear.  
- When dropped the Scales sidebar will update with x. 
- On delete the mark will disappear, but the scale will remain. 
- The container group will remain. 

Scenario 2:
- add a group 
- add a mark to the group
- drop data on each axis
- delete mark

**Expected outcome:** 
[...]

PR #387, opened 2016-07-14
Undo/Redo

arvind:
Track history of definitions of visualization components to enable undo/redo. Application state (inspectors, walkthroughs, hints, etc.) are not part of the history. 

- Reshape store to group together all "undoable" states, and use [redux-undo](https://github.com/omnidan/redux-undo) to add history tracking.

**This PR includes:**
- [x] Tests
- [x] functional comments

- Follow #331 and use undo/redo buttons in the toolbar to test undo/redo functionality. In particular,
  - The application UI (sidebars) should update and the Vega view should also update.
  - Binding a data field to a dropzone/channel should only register as a single element in the history state.
  - Continuous changes should be throttled/debounced and only register as a single historical state (e.g., dragging or resizing a mark, dragging a color picker, changing a number field input for axes).


arvind:
@kadamwhite: You may be interested in the approach I'm taking now. Besides regular history tracking functionality (storing, undo, redo), I have explicit `START_BATCH` and `END_BATCH` actions as well as a set of actions that get batched together implicitly when we cannot specify explicit start and end points to an interaction (e.g., dragging a mark, changing a color via the picker, etc.). The states produced by actions that fall within a batch do not get added to the history stack. If/when you have any spare cycles, I'd love a gut check on my logic in the new [historyReducer](https://github.com/vega/lyra/pull/387/files#diff-6e62469809082f2cb505b07240fc556f) (and absolutely no pressure/no worries if you don't!).

PR #397, opened 2016-08-19
Filter and Formula Data Transforms on DataTable

mattwchun:
This will add the feature to do Filter and Formula data transformations on the data in the DataTable component.

- filter data transforms in the DataTable
- formula data transforms in the DataTable

**This PR includes:**
- [ ] Tests
- [ ] functional comments

## 
## 
## 

…data transformations to the DataTable


src/js/components/pipelines/ExpressionTextbox.jsx
```diff
@@ -0,0 +1,43 @@
...
+      <Property name="expressionProp" type="autocomplete"   autoType="expr" dsId={id} {...props} onChange={this.filter}>
+      </Property>
+    );
+  }
+});
+
+module.exports = {
+  connected: connect(mapStateToProps, mapDispatchToProps)(ExpressionTextbox),
+  disconnected: ExpressionTextbox
+}
```
arvind: As we add more transforms, it's likely that each of them will have their own special inspector interface. So let's put these files under `src/js/components/pipelines/transforms` (and rename this particular file to `Filter.jsx` as being specific to the Filter UI?).

src/js/components/pipelines/FilterField.jsx
```diff
@@ -0,0 +1,47 @@
...
+        show = true,
+        dsId = props.dsId,
+        time = 10;
+
+    this.props.showExpressionTextbox(dsId, show, time);
+
+  },
+
+  render: function() {
+    return (<Icon onClick={this.showTextbox} glyph={assets.filter} width="10" height="10" />);
```
arvind: If you rebase to the latest `lyra2`, you'll see how I've introduced tooltips to these icons (e.g., with [`FieldType`](https://github.com/vega/lyra/commit/b7b364e1b396b63a4c9dd80d0f6af3c9011eac26#diff-767424686b9dcce6f942a1ca4d0823ae)). It would be good to add a tooltip to each of the new buttons you introduce as well.

src/js/reducers/datasetsReducer.js
```diff
@@ -33,8 +33,23 @@ function datasetsReducer(state, action) {
       field: action.field,
       order: action.order
     }));
+  }
+
+  if (action.type === ACTIONS.FILTER_DATASET) {
+    state = setIn(state, action.id + '._filter',
```
arvind: If we store the filter expression under `._filter` (and formula under `._formula`), the user can only specify a single filter and formula expression per dataset. Instead, perhaps we want to have more general `addTransform` and `updateTransform` actions? 

Both of these would take a `dsId` and a `def`, which is the Vega definition for a particular transform (e.g., `{"type": "filter", "test": ""}`). The reducer would then append the `def` to a dataset's `transform` array. `updateTransform` would also have to take an additional `idx` to identify which element in the `transform` array is being updated. 

This structure would also simplify the exporter logic as we don't need to do anything special to export the transforms. They'd be exported directly as part of a dataset.

src/js/reducers/datasetsReducer.js
```diff
@@ -33,8 +33,23 @@ function datasetsReducer(state, action) {
...
+      expression: action.expression
+    }));
+  }
 
+  if (action.type === ACTIONS.SHOW_EXPRESSION_TEXTBOX) {
+    state = setIn(state, action.id + '._expressionTextbox',
+    Immutable.fromJS({
+      show: action.show,
+      time: action.time
+    }));
```
arvind: I'm not sure this logic should be part of the store. Instead, because it's so specific to a particular React component, this may be one of the rare cases that we want to use the internal state of the component.

arvind:
Based on my last comment of moving the "show textbox" logic out of the store, here's how I imagine the flow operating. We'll use a running example of filtering by `Horsepower` from the cars dataset.

When a user clicks the filter icon, an `addTransform(dsId, {"type": "filter", "test": "datum.Horsepower "})` action is fired. Note, we pre-populate the expression with the field that was clicked on, so it should appear in the AutoComplete textbox. The reducer adds this transform to the corresponding dataset, and a `FilterTransform` component now appears above the `DataTable`. 

The `FilterTransform` component maintains an internal state that determines whether it's expanded (i.e., showing the autocomplete textbox) or contracted (showing just a button). An internal timer (`window.setTimeout`) keeps track of when to flip from expanded to contracted. When the user types in the textbox, the timer is cleared and restarts. This logic will likely be shared across all transform components, so we may want to create a parent `Transform` component within which we instantiate the particular "expanded" state. Here's a sketch:

**Transform.jsx**

``` js
getInitialState: function() {
  return {expanded: true};
},

timer: null,

resetTimer: function() {
  var that = this;
  window.clearTimeout(this.timer);
  this.timer = window.setTimeout(function() { 
    that.setState({expanded: false});
  }, 10000);
},

render: function() {
  return this.state.expanded ? this.props.children : (<div>...</div>);
}
```

**FilterTransform.jsx**

``` html
<Transform>
    <Property type="autocomplete" ... />
</Transform>
```

src/js/actions/datasetActions.js
```diff
@@ -47,6 +51,34 @@ function sortDataset(dsId, field, order) {
...
+
+function showExpressionTextbox(dsId, show, time) {
+  return {
+    type: SHOW_EXPRESSION_TEXTBOX,
+    id: dsId,
+    show: show,
+    time: time
+  };
+}
+
```
arvind: I don't think these two action creators are used anymore? Be sure to clean them up (along with associated variables and exports).

src/js/ctrl/export.js
```diff
@@ -80,6 +81,13 @@ exporter.dataset = function(state, internal, id) {
     spec.transform.push(sort);
   }
 
+  if (transforms) {
+    transforms.forEach(function(element, index, arr) {
+      spec.transform = spec.transform || [];
+      spec.transform.push(element);
+    });
+  }
+
```
arvind: If you store the transforms under `dataset.transforms` (rather than `._transforms`) you won't need this code anymore -- the transforms will get automatically exported when we call `.toJS()` earlier in this method.

src/js/components/pipelines/transforms/Inspector.jsx
```diff
@@ -0,0 +1,55 @@
...
+  var id = ownProps.dsId;
+  return {
+    transforms: getInVis(state, 'datasets.' + id + '._transforms')
+  };
+}
+
+var Inspector = React.createClass({
+  propTypes: {
+    dsId:  React.PropTypes.number,
+    transforms:  React.PropTypes.array
```
arvind: `transforms` should be an `ImmutableList` in the store (and marked here as a `React.PropTypes.instanceof(Immutable.List)`).

src/js/components/pipelines/transforms/Inspector.jsx
```diff
@@ -0,0 +1,55 @@
...
+  },
+
+  shouldComponentUpdate: function(nextProps, nextState) {
```
arvind: Because our store uses [ImmutableJS](https://facebook.github.io/immutable-js/), you don't need this method to test for deep equality in the passed in properties. Instead, ImmutableJS will produce a new ImmutableList for the `transforms` property if it is ever updated, which will have a new reference and thus trigger a re-render.

src/js/reducers/datasetsReducer.js
```diff
@@ -38,6 +39,41 @@ function datasetsReducer(state, action) {
...
+    if (!transforms) {
+      transforms = [];
+    }
+    var result = deepEquality.indexOf(action.transformSpec, transforms);
+    if (result === -1) {
+      // if transform doesn't exist already
+      transforms.push(action.transformSpec);
+    }
+
+    state = setIn(state, action.id + '._transforms', transforms);
```
arvind: Aha, part of the issues you're seeing is because `transforms` is a plain JavaScript array and then yes you're right, the reference to the array stays the same and so React never re-renders the component. Instead, as we do elsewhere in this file, you'll want to call `Immutable.fromJS` to convert a JavaScript array back to its equivalent ImmutableJS primitive _before_ putting it back in the store.

src/js/reducers/datasetsReducer.js
```diff
@@ -38,6 +39,41 @@ function datasetsReducer(state, action) {
     }));
   }
 
+  if (action.type === ACTIONS.FILTER_DATASET) {
```
arvind: This code path isn't necessary anymore.

src/js/reducers/datasetsReducer.js
```diff
@@ -38,6 +39,41 @@ function datasetsReducer(state, action) {
...
+
+    if (!transforms) {
+      transforms = [];
+    }
+    var index = action.specId;
+    if (index > -1) {
+      // if transform exist already
+      transforms[index] = action.newSpec;
+    }
+    state = setIn(state, action.id + '._transforms', transforms);
```
arvind: When editing, you can probably call `setIn` using the passed in index. For example, `return setIn(state, action.id + '.transforms.' + index, Immutable.fromJS(action.newSpec))`.

arvind:
Great work @mattwchun! You may want to take a look at the changes I made to this as some of them will be applicable to your next task. The commit diffs make it look like a more major change than it actually is. Besides renaming some of the actions and components to fit with our scheme more, and removing one level of nesting from the `transforms` component directory, the change flips how the transform components instantiate each other. 

Whereas previously you had an `ExpressionTextbox` component (now renamed to `TransformInspector`) that each transform's inspector called, I inverted this logic (i.e., `TransformInspector` creates a `Filter` or `Formula` inspector as needed). This allows us to share more common code in the `TransformInspector` (both the timer and connected action code) and pass that down to each individual transform via props. Keeps things simpler, with fewer connected components. 

Finally, the formula transform should also update the dataset's schema as necessary. In 927973092c579a1bb4bebddd330e210f5f9a748e, prior to merging this PR, I moved dataset schema definitions into the store. So updating the schema should be much easier now.

Issue #283, opened 2016-03-09
Switch from Dropdowns to Buttons for some text properties


kadamwhite:
### Description of requested feature

Lyra 1 supported button-based interaction for certain properties of text marks:
![image](https://cloud.githubusercontent.com/assets/442115/13646120/0a0a4262-e5fc-11e5-85a3-e7b29cf9b2a6.png)
![image](https://cloud.githubusercontent.com/assets/442115/13646125/0f5f0c2a-e5fc-11e5-82aa-03f641807e03.png)

These functions have been implemented in #281 using a select menu:

![image](https://cloud.githubusercontent.com/assets/442115/13646172/3d61dde6-e5fc-11e5-9b27-96db0fef8ad7.png)

We should create a property inspector component that permits setting these properties through buttons, rather than dropdowns, to improve the current Lyra 2 user experience and achieve parity with the Lyra 1 UI.
### Why this feature is important to you

Select menu dropdowns are a sub-par user experience for a number of reasons, and a toggle button (for bold or italic) or multiple graphical buttons (for e.g. alignment or baseline) more closely maintain the paradigm established by popular software like word processors.

nikhilkhanna:
So I've gotten this all implemented functionality-wise; however, Font Awesome does not have the images that the previous version of Lyra used for vertical alignment and from the old version of Lyra I could only find png versions of those assets, not the svg ones. Are there any other svg assets that would work well in place of the old vertical alignment assets?

arvind:
You might try https://thenounproject.com to see if there are icons there that match. We've used a handful of noun project icons, and have commented out the attribution text in the SVG (which we'll pull out and include in our "about" dialog at the end).

nikhilkhanna:
I've also noticed that, at least how it is now, setting the alignment of the text to, say "top" actually moves the text down and setting it to "bottom" moves the text up. Similarly setting alignment "left" moves the text to the right and setting alignment "right" moves the text to the left. Is this behavior as intended or should I correct this?

arvind:
Addressed w/#402. Re: how the text appears to shift, that's right -- the alignment (horizontal + vertical) determine how the (x, y) coordinates of a text mark position it. By default, with center+middle alignment, the (x, y) are the center of the text mark. Switching to "top" will make the text mark shift down since the (x, y) now determine the center-top coordinate. Nice work @nikhilkhanna!

PR #402, opened 2016-09-07
Switch from dropdowns to buttons for some text properties

nikhilkhanna:
This pull request addresses issue #283 to switch from dropdowns to buttons for some properties of text

- Change italic and bold text controls to use a toggle-able symbol as in a word processor
- Change alignment controls to use a selectable icon (as in a word processor)

- Ensure that text controls still function as intended (including double selecting controls)


src/js/components/inspectors/FormInputProperty.jsx
```diff
@@ -186,6 +188,48 @@ var FormInputProperty = React.createClass({
           </select>
         );
 
+      case 'toggle':
+        var currentVal = sg.get(props.signal);
```
arvind: I don't think you need `currentVal` here (and below). Instead, a `value` variable has already been declared and hooks into the local state management this component does.

src/js/components/inspectors/FormInputProperty.jsx
```diff
@@ -186,6 +188,48 @@ var FormInputProperty = React.createClass({
...
+        for (var i = 0; i < props.opts.length; i++) {
+          var val = props.opts[i];
```
arvind: Careful with `var` declarations within loops or if statements. JavaScript does not have block-level variable scoping. Instead, with Lyra, we prefer declaring all the variables a function uses at the very top. If you execute `npm run lint`, the linter should pick up this (and other) issues.

arvind:
Hi Nikhil, this looks pretty good to me. I've left you some minor comments in the code that should be easy to take care of.

The build was breaking previously because of a test dependency that changed the way its API was being exported. I fixed it in our master branch (f6a06ca) and rebased your's here. So you'll want to `git pull` first before making any additional changes.

arvind:
Merged! Thanks @nikhilkhanna :)

PR #414, opened 2018-11-28
Initial refactor of Pipeline factory to Immutable Record + TypeScript

arvind:
I think this PR roughly describes the changes we'll want to make across the factories and dependent files. Essentially, we use ImmutableJS's [Record types](https://facebook.github.io/immutable-js/docs/#/Record) which conveniently are factories themselves. The `RecordOf` type can then be used throughout the rest of the codebase to give us type safety.


src/js/util/dataset-utils.ts
```diff
@@ -59,7 +61,7 @@ function reset() {
  * @param {number} id - The ID of the dataset.
  * @returns {Array|string} An array of objects.
  */
-function input(id) {
+function input(id: number) {
```
jonathanzong, line 64: is it necessary to annotate the return value here and elsewhere?
i.e.
`function input(id: number): Array | string {`

src/js/util/dataset-utils.ts
```diff
@@ -88,15 +90,16 @@ function output(id) {
...
-  return promisify(dl.load)({url: url})
-    .then(function(data) {
-      return {data: data, pipeline: pipeline, dataset: dataset};
-    });
+function loadURL(url: string) {
+  const name = url.match(NAME_REGEX)[1];
+  const pipeline = Pipeline({name});
+  const dataset = Dataset(name, {url: url});
+
+  pipeline._id;
```
jonathanzong: does this line do anything?

arvind: Hah, no. Good catch. I think I typed this when I was checking VSCode's autocomplete capabilities with the `Record` types.

src/js/util/dataset-utils.ts
arvind: Good point! Yes, I think we should annotate the return values whenever possible. Will fix.

arvind: I decided not to add a type annotation for the return value because I'm not yet sure why the jsdoc marked it as `Array | string`. Seems like an opportunity for a cleaner design. Something to come back to!

PR #415, opened 2018-11-28
Change UndoRedo to TS.

alundgard:
* Change the toolbar/UndoRedo to typescript
* Change instances of require to import when referencing the changed file


src/js/components/toolbar/UndoRedo.tsx
```diff
@@ -0,0 +1,57 @@
+'use strict';
+// TODO Change from 'requires' to 'import'
+const historyActions = require('../../actions/historyActions');
+const undo = historyActions.undo;
+const redo = historyActions.redo;
+const assets = require('../../util/assets');
+const Icon = require('../Icon');
+
+import * as createReactClass from 'create-react-class';
+import * as propTypes from 'prop-types';
```
jonathanzong: this file appears not to use createReactClass or propTypes, so can probably remove these lines. (Also, we should generally be removing usages of createReactClass in favor of class .. extends React.Component.

src/js/components/toolbar/UndoRedo.tsx
```diff
@@ -0,0 +1,57 @@
...
+import * as React from 'react';
+import {connect} from 'react-redux';
+import {State} from '../../store/index';
```
jonathanzong: `import {State} from '../../store';` should just work

package.json
```diff
@@ -129,5 +129,10 @@
     "test:chrome": "npm run test -- --watch --browsers Chrome",
     "test:osx": "npm run test -- --browsers PhantomJS,Chrome,Firefox,Safari",
     "test:win": "npm run test -- --browsers PhantomJS,Chrome,Firefox,IE,IE9,IE10"
+  },
+  "dependencies": {
+    "@types/create-react-class": "^15.6.2",
+    "@types/react-redux": "^6.0.10",
+    "save": "^2.3.3"
```
jonathanzong: why was `save` added?

package.json
```diff
@@ -129,5 +129,10 @@
     "test:chrome": "npm run test -- --watch --browsers Chrome",
     "test:osx": "npm run test -- --browsers PhantomJS,Chrome,Firefox,Safari",
     "test:win": "npm run test -- --browsers PhantomJS,Chrome,Firefox,IE,IE9,IE10"
+  },
+  "dependencies": {
+    "@types/create-react-class": "^15.6.2",
```
jonathanzong: I think we shouldn't add this—usages of create-react-class should be replaced with class..extends rather than ported to TypeScript.

arvind: I just realized that `create-react-class` isn't actually in our `package.json`. Is this something that React provides automatically? I needed to manually add it `yarn add --dev create-react-class` when upgrading the `react-modal` dependency. Strange.

PR #416, opened 2018-11-29
Begin typing dataset actions and reducer

arvind:
Take a look at this after #414 has been merged.

In this branch, I'm starting to spec out what refactoring the redux part of the codebase looks like. In particular, here are the steps I've figured out so far:

1. When going from `.js` to `.ts`, large changes to the resultant `.ts` file can make it appear as though we deleted the `js` file and added a new `ts` file. An alternate strategy is to rename and then edit the files as two separate commits. Though this increases commit noise, we preserve the renaming history. This will give us diffs that are easier to parse. E.g., [`datasetActions.js -> datasetActions.ts`](https://github.com/vega/lyra/compare/as/typedDatasetActionsReducer?expand=1#diff-3a8bd0c6420d1651e6a9746894be3cb7).

2. Type the state by:
    1. Creating a Lyra type that extends the [vega-typings](https://github.com/vega/vega-typings) as necessary (e.g., [`LyraDataset`](https://github.com/vega/lyra/compare/as/typedDatasetActionsReducer?expand=1#diff-a7f6e053c00f56fffdcea0ce18bfebd4R30) extends Vega's `Data` via a type intersection). Lyra's state is a superset of a valid Vega specification. Lyra-specific properties are denoted by a leading underscore (e.g., `_id`). These properties are stripped out on export for a clean Vega specification.
    2. Creating the corresponding [Record factory and State type](https://github.com/vega/lyra/compare/as/typedDatasetActionsReducer?expand=1#diff-a7f6e053c00f56fffdcea0ce18bfebd4R37)

3. Convert basic action creators using the [typesafe-actions](https://github.com/piotrwitek/typesafe-actions) package. The [four](https://github.com/vega/lyra/commit/299df2e698a35a25091997441388e7fb63a43e0b#diff-2a3889c1790a8b9b8123d10eebf627a1R9) datasetActions I've refactored should give you a sense of how to type the `payload` and `meta` parts of an action. 

4. Follow and update references to these actions.

5. Update the corresponding reducer. In particular, this will involve:
    1. Using typesafe-action's `ActionType` and `getType` helpers.
    2. Transition away from the `set/setIn/get/getIn` helpers from our local `immutable-utils` and rely on the built-in one. Doing so gets us closer to more a type safe future.


arvind:
Some other thoughts working on this branch:

- The Vega typings may sometimes be too brittle for us. For instance, the [Data](https://github.com/vega/vega-typings/blob/master/types/spec/data.d.ts#L33) type is unioned inline rather than providing a separate type for each kind of dataset definition. As a result, the typescript compiler complains when we reference properties on [one of the sub-types](https://github.com/vega/lyra/compare/as/typedDatasetActionsReducer?expand=1#diff-94d024ffb1596e53aa8d06955aff9de1R45). In such cases, we should improve the source vega-typings, submit a PR, and `yarn link` locally until the improvements are released. For this specific case, we should break apart the `Data` type, and add a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) in vega-typings for us to use in Lyra.

- I'm going to mark this PR as a WIP until we convert the remaining actions for sorting and transformations (we can wait on the aggregate ones for a subsequent PR). As a result, this PR is a good next step for you two, once you get comfortable refactoring more of the React components. Might be a good candidate for pair programming? 
    - Convert sorting: extend `LyraDataset` with a `_sort` property, convert the action, update references, and then the reducer.
    - Convert transformations: Lyra supports two transformations currently: filter and formula. You'll first need to introduce typings for them in the factory before any follow-on refactoring. Unfortunately, the definitions for these two shifted from Vega 2 to Vega 3+. So my advice would be to implement them as they're defined in `vega-typings`, and then manually verify things work on the console as the Lyra interface will definitely break.

- `datasetsReducer.test.js` should be updated!

src/js/store/factory/Dataset.ts
```diff
@@ -27,15 +27,26 @@ export const Column = Record<ColumnDescription>({
 export type ColumnRecord = RecordOf<ColumnDescription>;
 
 export type Schema = Map<string, ColumnRecord>;
-export type LyraDataset = {
+
+// tslint:disable-next-line:class-name
+interface _LyraDataset {
```
jonathanzong: LyraDatasetProperties

src/js/components/pipelines/PipelineModal.tsx
```diff
@@ -41,8 +41,8 @@ function mapStateToProps(state: State, ownProps: OwnProps) {
 
 function mapDispatchToProps(dispatch, ownProps: OwnProps): DispatchProps {
   return {
-    addPipeline: function(pipeline, dataset, schema, values) {
-      dispatch(addPipeline(pipeline, dataset, schema, values));
+    addPipeline: function(pipeline, dataset, values) {
+      dispatch(addPipeline(pipeline, dataset, values));
     }
```
jonathanzong: mapDispatch

src/js/components/pipelines/PipelineModal.tsx
```diff
@@ -101,16 +99,15 @@ export class PipelineModal extends React.Component<OwnProps & DispatchProps, Pip
       'Successfully imported ' + url.match(NAME_REGEX)[0] + '!' : msg;
 
     dsUtils.loadURL(url)
-      .then(function(loaded) {
+      .then(function(loaded: {data: string, pipeline: PipelineRecord, dataset: DatasetRecord }) {
```
jonathanzong: name this type

src/js/actions/datasetActions.ts
```diff
@@ -16,27 +14,17 @@ export const deleteDataset = createStandardAction('DELETE_DATASET')<number, numb
 
 export const changeFieldMType = createStandardAction('CHANGE_FIELD_MTYPE')<{field: string, mtype: MType}, number>();
 
-// TODO: End of Arvind's typesafe datasetActions refactor.
+export const sortDataset = createStandardAction('SORT_DATASET')<{field: string, order: 'asc' | 'desc'}, number>();
```
jonathanzong: SortOrder

src/js/components/pipelines/transforms/SortIcon.tsx
```diff
@@ -35,7 +34,7 @@ function mapStateToProps(state: State, ownProps: OwnProps): StateProps {
 function mapDispatchToProps(dispatch: Dispatch, ownProps: OwnProps): DispatchProps {
```
jonathanzong: mapDispatch new syntax

src/js/constants/sortOrder.ts
```diff
@@ -0,0 +1,7 @@
+export const ASC = 'asc';
+export const DESC = 'desc';
+export type SortOrder = 'asc' | 'desc';
```
jonathanzong: make this an enum

src/js/actions/datasetActions.ts
```diff
@@ -17,27 +16,8 @@ export const changeFieldMType = createStandardAction('CHANGE_FIELD_MTYPE')<{fiel
 
 export const sortDataset = createStandardAction('SORT_DATASET')<{field: string, order: SortOrder}, number>();
```
jonathanzong: `Compare`
`CollectTransform`

src/js/actions/bindChannel/parseData.ts
```diff
@@ -37,14 +37,14 @@ module.exports = function(dispatch, state, parsed) {
 };
 
 function parseAggregate(dispatch, state, parsed, summary) {
-  const aggregate: AggregateTransform = summary.transform.find(function(tx) {
+  const aggregate: LyraAggregateTransform = summary.transform.find(function(tx) { // TODO(jzong) is this the right type? this file is scary
```
jonathanzong: remove todo

src/js/store/factory/Pipeline.ts
```diff
@@ -1,4 +1,7 @@
 import {Map, Record, RecordOf} from 'immutable';
+import {AggregateTransform} from 'vega-typings/types';
+
+export type LyraAggregateTransform = { _id: number } & AggregateTransform; // TODO(jzong) find an appropriate place to put this
```
jonathanzong: remove todo

arvind:
LGTM @jonathanzong!!

jonathanzong:
boooooooom

arvind:
💃 🕺 👯‍♂️ 👯 🎉

PR #424, opened 2019-01-24
Refactor pipelines

jonathanzong:
all files in components/pipelines (other than tests) refactored to es6/typescript


alundgard (approved):
Seems to be fine (compiles okay, no obvious errors). Remember to make a new commit for each filename change to preserve the diff. =)

arvind:
Looks like there's some overlap here with my approach in #416. I'd take a look at that PR to see if there are differences and/or to make sure we've caught everything for our refactoring (e.g., are we making use of vega-typings, so we know we're always producing valid Vega output).

PR #425, opened 2019-02-05
Al/inspectors

alundgard:
Refactor /inspectors. Add prop interfaces for state and dispatch.


src/js/components/inspectors/SpatialPreset.tsx
```diff
@@ -0,0 +1,113 @@
...
+  band: boolean,
+  group: boolean,
+  scale: any // TODO: propTypes.instanceOf(Immutable.Map),
+  name: string,
+  setPreset: any,
+  reset: any,
+  className: any
+}
+
+class BaseSpatialPreset extends React.Component<SpatialPresetProps>{
```
jonathanzong: `SpatialPresetProps` is currently redundant with `OwnProps`, `StateProps`, and `DispatchProps`.

Just as a reminder: the generic type `P` in `React.Component<P>` represents the prop types for the class. The props are split into `OwnProps`, which are regular props; `StateProps`, which are the outputs of mapStateToProps; and `DispatchProps`, which are the outputs of mapDispatchToProps. So in this case you want to:

1. Change the type in this line from `SpatialPresetProps` to `OwnProps & StateProps & DispatchProps`
2. Move fields that are present on `SpatialPresetProps` but not present in any of `OwnProps`, `StateProps`, and `DispatchProps` to `OwnProps`
3. Delete the definition for `SpatialPresetProps`

Please double check the other files for instances of this too :)

src/js/components/inspectors/Text.tsx
```diff
@@ -0,0 +1,137 @@
...
+        'properties.update.text.template', val));
+    }
+  };
+}
+
+interface TextInspectorProps {
+  primId: number,
+  primType: any, // TODO
+  autoVal: string,
+  dsId: any,
```
jonathanzong: `dsId` is a `number`

(it's usually referring to a dataset id)

jonathanzong (changes requested):
nice work! one class of changes to make and components should be good to go 🎉

src/js/components/inspectors/SpatialPreset.tsx
alundgard: 👍

src/js/components/inspectors/Text.tsx
alundgard: 👍

jonathanzong (approved):

PR #427, opened 2019-02-08
type all remaining factories, actions, reducers, and the store

jonathanzong:
So, this PR ballooned because of interlocking dependencies.

- Marks store, actions, and reducers: Area, Group, Line, Rect, Scene, Symbol, Text
- Parent Mark store, actions, reducer
- vegaActions, vegaReducer, introduce immutable store type called VegaReparse
- bindChannel/helperActions
- signalActions, signalsReducer, introduce immutable store type for Signals
- historyActions, historyReducer
- add types to the root store
- Hints, Inspector, and Walkthrough: refactor actions and reducers, introduce immutable store types
- fix a ton of runtime errors

This should more or less conclude our work on the store, actions, and reducers.

When running the head of this branch, things will render but nothing will work until we refactor bindChannel.


arvind:
Wowsa, fantastic work @jonathanzong

src/js/store/factory/marks/Area.ts
```diff
@@ -0,0 +1,86 @@
...
+  const y2 = propSg(id, 'area', 'y2');
+  const w = propSg(id, 'area', 'width');
+  const h = propSg(id, 'area', 'height');
+  const DELTA = sg.DELTA;
+  const DX = DELTA + '.x';
+  const DY = DELTA + '.y';
+  const streams: {[s: string]: OnEvent[];} = {};
+
+  streams[x] = [{
+    events: DELTA, update: test(at() + '||' + at('left'), x + '+' + DX, x)
```
arvind: `DELTA` refers to a signal name, and so should be wrapped in a SignalRef object, like so:

```js
  streams[x] = [{
    events: {signal: DELTA}, update: test(at() + '||' + at('left'), x + '+' + DX, x)
```

You may consider extracting it to the `DELTA` definition above, and update `DX` and `DY` instead.

src/js/store/factory/marks/Area.ts
```diff
@@ -0,0 +1,86 @@
+'use strict';
+
+import {Record, RecordOf} from 'immutable';
+import {AreaMark, OnEvent} from 'vega-typings';
+
+const anchorTarget = require('../../../util/anchor-target');
+const test = require('../../../util/test-if');
+const propSg = require('../../../util/prop-signal');
+
+export type LyraAreaMark = {
```
arvind, line 12: I wonder if it's worth using a more general `LyraMark` if the `_id` is the only field that's ever being added? I wonder if you can do something like `LyraMark<AreaMark>`?

src/js/store/factory/marks/Area.ts
```diff
@@ -53,38 +54,33 @@ Area.getHandleStreams = function(area): Map<string, OnEvent[]> {
   const DELTA = sg.DELTA;
   const DX = DELTA + '.x';
   const DY = DELTA + '.y';
-  const streams = {};
+  const streams: {[s: string]: OnEvent[];} = {};
```
arvind: Minor but might be worth extracting this type definition so that we're consistent across all marks?

src/js/store/factory/marks/Text.ts
```diff
@@ -40,57 +42,30 @@ function Text() {
...
+    {events: DELTA, update: test(at('left') + '&&' + at('top'), fontSize + '-' + DY, fontSize)},
+    {events: DELTA, update: test(at('right') + '&&' + at('bottom'), fontSize + '+' + DY, fontSize)}
   ];
   return streams;
 };
-
-/**
- * @property alignments {string[]} Valid align properties for vega text marks
- */
-Text.alignments = ['left', 'center', 'right'];
```
arvind: I'm curious why we're deleting all of these? I think we used these to populate drop downs in the UI. Are we getting these values some other way (e.g., maybe directly from the typings somehow?).

src/js/store/factory/marks/Group.ts
```diff
@@ -1,54 +1,33 @@
...
-    Rect = require('./Rect'); // Visually, groups are similar to Rects.
+import {Record, RecordOf} from 'immutable';
+import {GroupMark} from 'vega-typings';
+import {Rect} from './Rect';
 
+const dl = require('datalib');
+const getInVis = require('../../../util/immutable-utils').getInVis;
 
 export type LyraGroupMark = {
-  _id: number
```
arvind: Why no `_id` here? I guess would be moot with a more general `LyraMark`.

src/js/store/factory/marks/Group.ts
```diff
@@ -59,10 +38,7 @@ function Group() {
         y2: {_disabled: true}
       })
     }
-  };
+  })(values);
 }
 
-Group.getHandleStreams = Rect.getHandleStreams;
```
arvind: Do we want to import Rect's `handleStreams` and then re-export it?

src/js/store/factory/marks/Scene.ts
```diff
@@ -1,25 +1,43 @@
...
-}
+const dl = require('datalib');
 
-Scene.getHandleStreams = function() {
-  return {};
-};
+// TODO(jzong): how to keep this in sync with GroupMark in vega-typings?
+export interface LyraSceneMark extends BaseMark, Scope, Encodable<GroupEncodeEntry> {
+  type: 'scene'
+  from?: FromFacet;
```
arvind: Lyra's `Scene` corresponds to the [top-level of a Vega specification](https://vega.github.io/vega/docs/specification/) which corresponds to the top-level group mark in the resulting scenegraph. As a result, it shouldn't have a `from` property.

arvind: I'm also confused by the `type: "scene"` but I see it there in the old codebase, so perhaps it means something and some other part of the codebase does something with it? We should mark it as a todo (or perhaps file an issue which will be easier to track) to follow up though.

src/js/store/factory/Mark.ts
```diff
@@ -64,27 +76,31 @@ function Mark(type, props) {
...
+//     x = arguments[i];
+//     for (key in x) {
+//       obj[key] = x[key];
+//       if (x[key] === undefined) {
+//         delete obj[key];
+//       }
+//     }
+//   }
+//   return obj;
+// }
```
arvind: Delete?

src/js/store/factory/marks/Group.ts
```diff
@@ -6,15 +6,17 @@ const dl = require('datalib');
 const getInVis = require('../../../util/immutable-utils').getInVis;
 
 export type LyraGroupMark = {
+  _id: number,
```
arvind: Aha, it came back!

arvind (commented):
First batch of comments up to and including 9dea975.

src/js/actions/markActions.ts
```diff
@@ -1,58 +1,51 @@
...
-  // We pull in all of the mark constructors purely to access their static
-  // `.getHandleStreams` and `.defaultProperties` methods
-  var props = dl.extend({
-    _id: markProps._id || counter.global(),
-  }, markProps);
+// TODO: use the vega-lite unit type, probably defined here:
+// https://github.com/vega/vega-lite/blob/d61deb9e4a2312114ce74945af65fc165309899a/src/spec/unit.ts
+// may require modification of vega-lite typings
+// import { GenericUnitSpec } from 'vega-lite/src/spec/unit';
+type VegaLiteUnit = any;
```
arvind: Yup, that's the one (or, a non-commit specific version here: https://github.com/vega/vega-lite/blob/master/src/spec/unit.ts). It'll likely require a Lyra-specific wrapper like our mark factories.

src/js/reducers/sceneReducer.ts
```diff
@@ -1,20 +1,18 @@
...
   if (typeof state === 'undefined') {
-    return Immutable.Map();
+    return Map();
   }
 
-  if (action.type === actions.CREATE_SCENE) {
-    return state.set('id', action.id);
+  if (action.type === getType(sceneActions.createScene)) {
+    // TODO(jzong) figure out what this line was supposed to do
+    // return state.set('id', action.id);
```
arvind: Because a "scene" is just a fancy group mark, I think what this did was map the "scene" state in the store to a particular mark id associated with the scene.

src/js/actions/markActions.ts
```diff
@@ -16,7 +16,7 @@ type VegaLiteUnit = object;
 
 export const addMark = createStandardAction('ADD_MARK').map((record: MarkRecord) => {
   const id: number = record._id || counter.global();
-  record = record.merge({_id: id});
+  record = (record as any).merge({_id: id}); // TODO(jzong) typescript barfs when calling merge on union record types
```
arvind: Classic TypeScript.

src/js/actions/Names.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+// TODO(jzong) eventually delete this file
```
arvind, line 3: 🎉

src/js/actions/index.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+// TODO(jzong) eventually delete this file
```
arvind, line 3: 🎉

src/js/components/hints/symbol-hint.tsx
```diff
@@ -1,38 +1,40 @@
...
 import * as React from 'react';
 import {connect} from 'react-redux';
 import { Dispatch } from 'redux';
+import {updateMarkProperty} from '../../actions/markActions';
 import {State} from '../../store';
-interface SymbolHintsProps {
-    selectedId: number,
-    updateProperty: (id: number, property: any, value: any) => void // TODO: fix 'any' type
+
+const lookup = require('../../ctrl').lookup; // TODO(jzong) ??? what is this? doesn't exist?
```
arvind, line 7: `Hints` were the newest most experimental part of the `lyra2` codebase. I wouldn't be surprised if a bunch of this code is super incomplete.

src/js/actions/signalActions.ts
```diff
@@ -13,13 +15,7 @@ var ns = require('../util/ns'),
...
  * @returns {Object} An action object
  */
-function initSignal(signal, value) {
-  return {
-    type: INIT_SIGNAL,
-    signal: ns(signal),
-    value: value
-  };
-}
+export const initSignal = createStandardAction('INIT_SIGNAL')<any, string>();
```
arvind: Like the other action files, could you add a comment to these lines to indicate what the two parameters are. E.g., I'm thinking `// initial value, signal name`

src/js/ctrl/signals/defaults.ts
```diff
@@ -1,87 +1,98 @@
-'use strict';
-var dl = require('datalib'),
-    ns = require('../../util/ns'),
-    signals = {};
+// TODO(jzong): this probably should be moved to store/factory
```
arvind: I'm not so sure. This file defines the Vega spec for the signals we need. The store holds the _values_ for the actual signals (i.e., what gets returned if you call Vega's view api `view.signal('name')`).

src/js/ctrl/signals/defaults.ts
```diff
@@ -1,87 +1,98 @@
...
-  ],
-  _idx: 3
-};
+export type LyraSignal = {
+  _idx: number
+} & VegaSignal;
+export const Signal = Record<LyraSignal>({
+  _idx: null,
+  name: null
+});
```
arvind: See above comment. Because the definitions here aren't to be embedded into the store, we don't need to wrap it in a `Record` and can work with the `LyraSignal` directly.

src/js/ctrl/signals/defaults.ts
```diff
@@ -1,87 +1,98 @@
...
-    {type: 'mousedown', expr: "eventItem() && eventItem().cursor || 'default'"},
-    {type: 'mouseup', expr: "'default'"}
-  ]
-};
+export const defaultSignalState: SignalState = Map({
+  SELECTED: Signal({
+    name: SELECTED,
+    value: {mark: {}},
+    on: [
+      {events: 'mousedown[eventItem().mark && eventItem().mark.name &&' +
```
arvind: These should be updated to the new [event functions](https://vega.github.io/vega/docs/expressions/#event-functions). For example, it's just `item()` now instead of `eventItem()`.

src/js/actions/signalActions.ts
```diff
@@ -1,12 +1,5 @@
 import {createStandardAction} from 'typesafe-actions';
-import {OnEvent} from 'vega-typings/types';
-
-const ns = require('../util/ns');
-
-    INIT_SIGNAL = 'INIT_SIGNAL',
-    SET_SIGNAL  = 'SET_SIGNAL',
-    SET_SIGNAL_STREAMS = 'SET_SIGNAL_STREAMS',
-    UNSET_SIGNAL = 'UNSET_SIGNAL';
+import {OnEvent, SignalValue} from 'vega-typings/types';
```
arvind, line 2: I'm curious about why we need to import from `vega-typings/types` rather than just `vega-typings` (or even `vega`)?

src/js/actions/historyActions.ts
```diff
@@ -1,49 +1,22 @@
...
-  CLEAR_HISTORY: CLEAR_HISTORY,
-  START_BATCH:  START_BATCH,
-  END_BATCH: END_BATCH,
+  UNDO: 'UNDO',
+  REDO: 'REDO',
+  JUMP_TO_FUTURE: 'JUMP_TO_FUTURE',
+  JUMP_TO_PAST: 'JUMP_TO_PAST',
+  CLEAR_HISTORY: 'CLEAR_HISTORY',
+  START_BATCH:  'START_BATCH',
+  END_BATCH: 'END_BATCH',
```
arvind: Do we still need to export these names?

src/js/reducers/historyReducer.ts
```diff
@@ -1,30 +1,36 @@
...
+const IMPLICIT_BATCH = [getType(signalActions.setSignal), getType(guideActions.updateGuideProperty)];
+const BATCH_INTERVAL = 500;  // ms to identify new batch for same implicit action.
+let batch = 0;
+let prevAction;
+let prevTime;
+
+export function undoable(reducer) {
+  return function(state: VisState, action: AnyAction): VisState {
     if (typeof state === 'undefined') {
       state = create([], reducer(undefined, {}), []);
```
arvind: I'm seeing some type inconsistencies in this reducer that I just want to double check. `past`, `present`, and `future` should all be arrays (and thus use `.length`) or Immutable `List`s (and use `.size`). Looking through this code, it seems like we might be mixing between these two types. I'm fine with either choice of collection.

src/js/reducers/inspectorReducer.ts
```diff
@@ -1,93 +1,67 @@
...
-    getIn = immutils.getIn,
-    setIn = immutils.setIn,
-    ACTIONS = require('../actions/Names');
-
-function expandLayers(state, layerIds) {
-  return layerIds.reduce(function(newState, layerId) {
-    return setIn(newState, 'encodings.expandedLayers.' + layerId, true);
+function expandLayers(state: InspectorRecord, layerIds: number[]): InspectorRecord {
+  return layerIds.reduce(function(newState: InspectorRecord, layerId: number) {
+    return newState.setIn(['encodings', 'expandedLayers', layerId], true);
```
arvind: Do we expect `layerId` to be a string or number? In the old code, `layerId` would get coerced to a string. Shouldn't matter so long as we're consistent.

src/js/actions/walkthroughActions.ts
```diff
@@ -1,46 +1,14 @@
...
-  SET_WALKTHROUGH_ON: SET_WALKTHROUGH_ON,
+export const setActiveStep = createStandardAction('SET_ACTIVE_STEP')<number>();
+export const setActiveWalkthrough = createStandardAction('SET_ACTIVE_WALKTHROUGH')<WalkthroughName>();
 
-  // Action Creators
-  setWalkthrough: setWalkthrough,
-  setActiveStep: setActiveStep,
-  setActiveWalkthrough: setActiveWalkthrough
-};
+// TODO(jzong): unused, need to clarify what the intent was
```
arvind, line 13: Like hints, walkthroughs was super new/experimental code. So very very unfinished.

arvind (approved):
This is an incredible effort @jonathanzong! I just finished my review, and left mostly minor comments throughout that would be good for you to double check prior to merging. The only place that I think we should double check things is with the historyReducer/state, to make sure we're being consistent about arrays/immutable Lists.

src/js/reducers/historyReducer.ts
jonathanzong: great catch! this got fixed in a later commit, where i ended up going with immutable lists.
 https://github.com/vega/lyra/pull/427/commits/17a77abb1711b96c5c1e300c4d671c0d0e63cf1c#diff-2255b1a147d62cf636821d5ae360a49c

src/js/actions/signalActions.ts
jonathanzong: all of the above will work, i think i just let vscode autocomplete pick every time. i will make it consistent

src/js/store/factory/marks/Area.ts
jonathanzong: filed an issue https://github.com/vega/lyra/issues/436

basically, group is the only outlier

src/js/store/factory/marks/Area.ts
jonathanzong: done in https://github.com/vega/lyra/pull/427/commits/881c1fa64c836a85391926033c5166a5a5c96e47

src/js/store/factory/marks/Area.ts
jonathanzong: Inferring from `ctrl/signals`, it looks like DELTA is currently a `SignalRecord` from the store. I've typed accordingly in https://github.com/vega/lyra/pull/427/commits/91d091cfab9b4a281714965f2e45a6191d044508 but @rneogy and I should revisit together when `ctrl/signals` gets converted to typescript.

src/js/store/factory/marks/Text.ts
jonathanzong: thank you for catching this, it was an overzealous delete. i added these back with vega types here: https://github.com/vega/lyra/pull/427/commits/6d9c1aca476bf67113b7cc1ae5308c88937f45c2

src/js/actions/markActions.ts
jonathanzong: Great, I've introduced `type VegaLiteUnit = UnitSpec` in https://github.com/vega/lyra/pull/427/commits/03e476da6f6647e6a1d1716cbaabf1f21d83ce59. fyi @rneogy this action, which now has a type, is dispatched in `bindChannel`

src/js/ctrl/signals/defaults.ts
jonathanzong: done in https://github.com/vega/lyra/pull/427/commits/5b6ce971270944c0ea2e7dd3eb59e089c46c55b6

src/js/store/factory/marks/Scene.ts
jonathanzong: filed issue https://github.com/vega/lyra/issues/434

arvind:
🎉🎉🎉

PR #429, opened 2019-02-14
Rn/typed guide actions reducer

rneogy:
Refactored Guide actions, reducers, and store, along with dependent files.


jonathanzong:
👍 🔥 🔥 🔥

src/js/components/encodings/GuideList.tsx
```diff
@@ -79,7 +81,7 @@ class GuideList extends React.Component<OwnProps & StateProps & DispatchProps> {
 
         {props.guides.map(function(guide) {
           const guideId = guide.get('_id');
-          const scaleId = guide.get('scale') || guide.get(guide.get('_type'));
+          const scaleId = guide.get('scale'); // || guide.get(guide.get('type')); TODO(rneogy) figure out if we need this?
```
arvind, line 84: The commented portion is definitely needed -- see the difference between specifying [axes](https://vega.github.io/vega/docs/axes/) are specified (using the `scale` property) and [legends](https://vega.github.io/vega/docs/legends/) (which use `fill`, `shape`, etc. depending on the legend type). That's why it's `_type` (i.e., a Lyra-specific property we added).

PR #437, opened 2019-04-01
bump vega typings version

jonathanzong:
upgrade vega-typings to latest so we no longer have to yarn link, fix the compile issues


rneogy (approved):
lgtm!

PR #438, opened 2019-04-02
fix signal runtime error

jonathanzong:
This PR fixes some default signal logic that got lost in the mega-PR.

- add export statement to some variables in `store/factory/Signal`
- refactor `ctrl/signal`
- update references, mostly in mark `getHandleStream` functions


src/js/ctrl/signals.ts
```diff
@@ -66,7 +62,7 @@ api.isDefault = isDefault;
  * @returns {*} The value of the signal
  */
 api.get = function(name) {
-  var ctrl = require('../'),
+  var ctrl = require('./'),
```
rneogy, line 65: Should probably refactor to use const/let instead of var while you're upgrading to typescript.

rneogy: Though if you're planning on rewriting this file anyways, I guess that's okay - I'll give the approval since everything else looks good!

rneogy (approved):
Looks good for now!

src/js/ctrl/signals.ts
rneogy: Although wait this line itself is really weird? Why are you requiring yourself?

jonathanzong: yeah this folder was kinda wacky -- it's requiring the index in that directory

PR #478, opened 2020-09-17
Esefah/legend inspector

### LINKED ISSUE #477: Legend Inspector
The controls like title and title font size are not working for the legend inspector.

NanaKwame:
Fixes #477. 
- Updated syntax in Legend.tsx to reflect new Vega syntax. Mainly properties -> encode. and added .value to the property names.
- Added change listeners to the inputs.
- Fixed the "this" bug in MoreProperties component.
- Fixed use of setIn in guideReducer to reflect updated use of how to pass property names.
- Added missing properties to legend record.

Submitting for code review. 

Only issue currently is that I am getting this warning:
Warning: A component is changing an uncontrolled input of type range to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component.

Looking into this.

### DIFF
--- a/src/js/components/inspectors/MoreProperties.tsx
+++ b/src/js/components/inspectors/MoreProperties.tsx
@@ -37,25 +37,25 @@ class BaseMoreProperties extends React.Component<MorePropsProps, MorePropsState>
       moreLink = (
         <div>
           <h3 className='show-more-props more-props-label'
-            onClick={this.handleClick} style={style1}>
+            onClick={(e) => this.handleClick(e)} style={style1}>
             + {props.label}</h3>
 
           <h3 className='hide-more-props more-props-label'
-            onClick={this.handleClick} style={style0}>
+            onClick={(e) => this.handleClick(e)} style={style0}>
             — {props.label}</h3>
         </div>
       );
     } else {
       moreLink = (
         <a className='show-more-props more-props-label' href='#'
-          onClick={this.handleClick} style={style1}>
+          onClick={(e) => this.handleClick(e)} style={style1}>
           + More {props.label} Properties
         </a>
       );
 
       fewerLink = (
         <a className='hide-more-props more-props-label' href='#'
-          style={style0} onClick={this.handleClick}>
+          style={style0} onClick={(e) => this.handleClick(e)}>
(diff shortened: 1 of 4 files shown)


NanaKwame:
Oh also default values aren't being reflected when you initially open the inspector. Another thing I am looking into.

src/js/components/inspectors/MoreProperties.tsx
```diff
@@ -37,25 +37,25 @@ class BaseMoreProperties extends React.Component<MorePropsProps, MorePropsState>
       moreLink = (
         <div>
           <h3 className='show-more-props more-props-label'
-            onClick={this.handleClick} style={style1}>
+            onClick={(e) => this.handleClick(e)} style={style1}>
```
ktbacher, line 40: Just a comment because I edited this file too, instead of changing all of the 'this.handleClick', I found you can just add 'this.handleClick = this.handleClick.bind(this);' to the constructor and these will work as is

EbenezerSefah: Yeah. So I chose this approach because it was more consistent with InteractionInputType.

jonathanzong:
This is really fantastic work so far. You've done a great job of connecting the inspector to the store, so now changes in the inspector are reflected in the visualization, which is exactly where we wanted to get started.

A next step we'll want to accomplish is making sure the inspector:
1) reads the initial values shown in the inspector from the store. For example, when I create a symbol and bind something to `fill`, it makes a new legend with `orient: right` by default, but the inspector initially shows `orient: left`
2) shows only the relevant properties to the legend. If you bind something to `fill` and that generates a legend, the legend inspector only needs to show `fill` and not the other un-bound stuff like opacity, etc.
3) shows the user when properties are bound to fields. If you make a symbol mark and then bind a field to `fill`, in the symbol inspector it will show the name of the field instead of a color picker. This indicates that the value is bound, rather than user-defined. you can use the symbol inspector as an example to make sure this happens in the legends inspector too.
4) shows all relevant properties for all marks. For example, create a symbol mark and then bind something to `size`. Size is a property specific to the symbol mark, but it should show up in the inspector.

PR #485, opened 2020-09-21
Guide Initial Value Fix

### LINKED ISSUE #483: Guides - read values from store
This is really fantastic work so far. You've done a great job of connecting the inspector to the store, so now changes in the inspector are reflected in the visualization, which is exactly where we wanted to get started.

A next step we'll want to accomplish is making sure the inspector:
1) reads the initial values shown in the inspector from the store. For example, when I create a symbol and bind something to `fill`, it makes a new legend with `orient: right` by default, but the inspector initially shows `orient: left`
2) has a section at the top showing the relevant scale name and the property/field binding. If you make a symbol mark and then bind a field to `fill`, in the symbol inspector it will show the name of the field instead of a color picker. This indicates that the value is bound, rather than user-defined. you can use the symbol inspector as an example to add this to the top section of the legends inspector

(edited after 9/24 meeting)

_Originally posted by @jonathanzong in https://github.com/vega/lyra/pull/478#issuecomment-695106613_

NanaKwame:
So far this PR addresses:
1. the issue of picking up default values. I updated the .ts files to reflect the correct default values and these are now working correctly in the inspector.
2. Fixed a null/0 bug in Property.tsx

### DIFF
--- a/src/js/components/inspectors/Property.tsx
+++ b/src/js/components/inspectors/Property.tsx
@@ -76,7 +76,7 @@ function mapStateToProps(reduxState: State, ownProps: OwnProps): StateProps {
   return {
     group:  getIn(state, path + '.group'),
     signal: getIn(state, path + '.signal'),
-    value:  value ? value : ownProps.value,
+    value:  value != null ? value : ownProps.value,
     field:  field,
     scale:  scale,
     srcField:  dsId && field ?
--- a/src/js/components/inspectors/Property.tsx
+++ b/src/js/components/inspectors/Property.tsx
@@ -76,7 +76,7 @@ function mapStateToProps(reduxState: State, ownProps: OwnProps): StateProps {
   return {
     group:  getIn(state, path + '.group'),
     signal: getIn(state, path + '.signal'),
-    value:  value != null ? value : ownProps.value,
+    value:  value !== null ? value : ownProps.value,
     field:  field,
     scale:  scale,
     srcField:  dsId && field ?
--- a/src/js/components/inspectors/Guide.tsx
+++ b/src/js/components/inspectors/Guide.tsx
@@ -1,5 +1,4 @@
 'use strict';
-const getInVis = require('../../util/immutable-utils').getInVis;
 
 import * as React from 'react';
 import {connect} from 'react-redux';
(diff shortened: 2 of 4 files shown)


src/js/components/inspectors/Property.tsx
```diff
@@ -76,7 +76,7 @@ function mapStateToProps(reduxState: State, ownProps: OwnProps): StateProps {
   return {
     group:  getIn(state, path + '.group'),
     signal: getIn(state, path + '.signal'),
-    value:  value ? value : ownProps.value,
+    value:  value != null ? value : ownProps.value,
```
jonathanzong: should this be `!==`?

jonathanzong:
lgtm other than the tiniest comment

src/js/components/inspectors/Property.tsx
NanaKwame: are we guaranteed it will never be undefined? Decided to make it != to deal with undefined but I can change it to !== if that isn't an issue.

NanaKwame:
I was planning on fixing the other issues in this PR. Should I wait till have have those fixes or should this be merged and a new issue created for the other fixes. lmk

src/js/components/inspectors/Property.tsx
NanaKwame: actually nvm will make the change

ktbacher:
> I was planning on fixing the other issues in this PR. Should I wait till have have those fixes or should this be merged and a new issue created for the other fixes. lmk

I think the initial value issue is the main goal with this PR and then the other point in issue #483 could be a separate PR, but was there something else you were thinking for this PR?

src/js/components/inspectors/Property.tsx
jonathanzong: what about (edit:`value || value === undefined`) ? does that do the same thing? just thinking about how to make the intent clear

NanaKwame: Sorry I am late to this. So the issue is that value can be 0 so that wouldn't do the same thing. Maybe we don't actually have to worry about undefined so I will change this to !== then.

NanaKwame:
> > I was planning on fixing the other issues in this PR. Should I wait till have have those fixes or should this be merged and a new issue created for the other fixes. lmk
> 
> I think the initial value issue is the main goal with this PR and then the other point in issue #483 could be a separate PR, but was there something else you were thinking for this PR?

Sorry I am getting to this late. Yeah that is a good point. I will add the change in label here and a fix for Jonathan's feedback and merge this. I will create a separate issue for the other things.

PR #487, opened 2020-09-22
Area Orientation selector

ktbacher:
Original purpose:  Expose Vega orientation property to Lyra user in area inspector.  Decided not to add this feature.  The components are there and commented out, in case we decide to add it in later.

- Commented out orientation selector for users to select either 'horizontal' or 'vertical' orientation.
- Added initial values to area mark
- Used <extendProperty> as opposed to hard-coded y and y2 to match rect component
- Switches which properties are editable (e.g. x2, y2)  in the area inspector if orientation is switched (leaving it set to vertical though)

### DIFF
--- a/src/js/components/inspectors/Area.tsx
+++ b/src/js/components/inspectors/Area.tsx
@@ -1,10 +1,12 @@
 'use strict';
 const INTERPOLATE = require('../../constants/interpolate');
+const getInVis = require('../../util/immutable-utils').getInVis;
 
 import * as React from 'react';
 import {connect} from 'react-redux';
 import {PrimType} from '../../constants/primTypes';
 import {Property} from './Property';
+import {State} from '../../store';
 
 const Area = {
   'ORIENT': [
@@ -12,12 +14,22 @@ const Area = {
     'horizontal'
 ]};
 
-interface AreaProps {
+interface OwnProps {
   primId: number,
   primType: PrimType
 }
 
-class BaseArea extends React.Component<AreaProps> {
+interface StateProps {
+  orient: string;
+}
+
+function mapStateToProps(reduxState: State, ownProps: OwnProps): StateProps {
+  const primId = ownProps.primId;
+  return {
+    orient: getInVis(reduxState, `signals.lyra_area_${primId}_orient.value`)
+  };
+}
+class BaseArea extends React.Component<OwnProps & StateProps> {
   public render() {
     const props = this.props;
     return (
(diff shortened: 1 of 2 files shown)


ktbacher:
Clicking the checkbox on the ExtendProperty component gives this error message:
`Warning: A component is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components`

NanaKwame:
Something that fixed that error for me was to specify a default value in the .ts for whatever property was causing that error.

src/js/components/inspectors/Area.tsx
```diff
@@ -1,39 +1,68 @@
...
+              <h3>Y Position</h3>
+              <ExtentProperty exType='y' {...props} />
 
-          <Property name='y2' label='End' type='number' droppable={true} {...props} />
-        </div>
+            </div>
+          :  <Property name='y' type='number' droppable={true} {...props}>
+                <h3 className='label'>Y Position</h3>
+              </Property>
+        }
```
jonathanzong: a formatting suggestion for this part, to get the divider lines to show up in the right places (they're attached to the `property-group` class)

```
        <div className='property-group'>
          <h3>X Position</h3>
          {props.orient === 'vertical' ?
              <Property name='x' label='X' type='number' droppable={true} {...props} />
            :
              <ExtentProperty exType='x' {...props} />
          }
        </div>

        <div className='property-group'>
          <h3>Y Position</h3>
          {props.orient === 'vertical' ?
              <ExtentProperty exType='y' {...props} />
            :
              <Property name='y' label='Y' type='number' droppable={true} {...props} />
          }
        </div>
```

src/js/components/inspectors/Area.tsx
```diff
@@ -1,39 +1,68 @@
...
 
-class BaseArea extends React.Component<AreaProps> {
+interface StateProps {
+  orient: string;
+}
+
+function mapStateToProps(reduxState: State, ownProps: OwnProps): StateProps {
+  const primId = ownProps.primId;
+  return {
+    orient: getInVis(reduxState, `signals.lyra_area_${primId}_orient.value`)
```
jonathanzong: i should have said something about this, but we're trying to move away from the `getInVis` helper function in favor of this syntax:

```
reduxState.getIn(['vis', 'present', 'signals', `lyra_area_${primId}_orient`, 'value'])
```

and i think it would be even better not to hardcode the signal name, just in case something about that changes down the line. so:

```
const orientSignal = reduxState.getIn(['vis', 'present', 'marks', str(primId), 'encode', 'update', 'orient', 'signal']);
return {
  orient: reduxState.getIn(['vis', 'present', 'signals', orientSignal, 'value'])
}
```

and if this fails, it'll be easier to tell right away what's going wrong.

PR #488, opened 2020-09-25
Default axis title and orient options

ktbacher:
Populate the default axis title from the parsed Vega def

- Select the parsed Vega axis def that corresponds to the data-bound axis
- Restricts axis orientation options to top and bottom for x axis and left and right for y axis

### DIFF
--- a/src/js/actions/bindChannel/parseGuides.ts
+++ b/src/js/actions/bindChannel/parseGuides.ts
@@ -123,11 +123,11 @@ function findOrCreateAxis(dispatch: ThunkDispatch<State, any, any>, state: State
   if (count < 2) {
     let axis = Guide(GuideType.Axis, parsed.channel, scaleId) as AxisRecord;
     axis = axis.mergeDeep({
-      title: def[0].title,
+      title: def[1].title,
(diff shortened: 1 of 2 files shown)


ktbacher:
I feel like there should be a better way to verify this is the correct axis.  From inspection, seems like def[0] is always the base layer/canvas axis, while def[1] corresponds to the first added axis bound to data, but I'm not sure this is always the case.

src/js/actions/bindChannel/parseGuides.ts
```diff
@@ -123,11 +123,11 @@ function findOrCreateAxis(dispatch: ThunkDispatch<State, any, any>, state: State
   if (count < 2) {
     let axis = Guide(GuideType.Axis, parsed.channel, scaleId) as AxisRecord;
     axis = axis.mergeDeep({
-      title: def[0].title,
+      title: def[1].title,
```
jonathanzong, line 126: const titleDef = def.find(x => x.title);

jonathanzong:
I'm gunna merge this and make a new issue for the parseGuides stuff

PR #496, opened 2020-10-06
Scale Inspector

NanaKwame:
Goal is to design the scale inspector.

Current state:
- first pass at showing just the fields we have access to


NanaKwame:
As of now only the first 4 fields show something. Need to figure out how to get more fields to get populated.

src/js/components/inspectors/Scale.tsx
```diff
@@ -20,22 +21,41 @@ interface StateProps {
...
     return (
       <div>
         <div className='property-group'>
           <h3 className='label'>Placeholder</h3>
           <ul>
             <li>name: {scale.get('name')}</li>
             <li>type: {scale.get('type')}</li>
-            {/* <li>range: {scale.get('range')}</li> TODO */}
+
+            <Property name='type' label='Type' type='select' opts={typeOpts} {...props} />
```
ktbacher: I'm not sure how the onChange works here, doesn't seem to be updating the property

src/js/components/inspectors/FormInputProperty.tsx
```diff
@@ -189,6 +189,9 @@ class BaseFormInputProperty extends React.Component<OwnProps & StateProps & Disp
         );
 
       case 'select':
+        if (value.scheme) {
```
ktbacher: Might be a little hacky, open to suggestions on other ways to handle this.  `value = {scheme: "someColorScheme"}` otherwise and threw off the selector options

src/js/components/inspectors/Scale.tsx
```diff
@@ -18,29 +21,103 @@ interface StateProps {
...
+    const target = evt.target;
+    const property = target.name;
+
+    let value = (target.type === 'checkbox') ? target.checked : target.value;
+    // Parse number or keep string around.
+    value = value === '' || isNaN(+value) ? value : +value;
+    this.props.updateScaleProperty(scaleId, property, value);
+  };
+
+  public handleRangeColor(evt) {
```
ktbacher: not sure if there is a way to handle this in the other `handleChange()` function but need to pass an object for the color scheme value

NanaKwame: gotcha. moved it to handlechange. lmk if it looks ok.

src/js/store/factory/Scale.ts
```diff
@@ -16,7 +16,8 @@ export type LyraScale = {
    */
   _origName: string;
   _domain: any[],
-  _range: any[]
+  _range: any[],
+  nice: any
```
jonathanzong: as far as i know, no need to add this here. it's defined in the definitions of e.g. `ThresholdScale`, `QuantizeScale`, etc, which are part of `RangeScale`, and included when we use `& RangeScale` to sort of extend `RangeScale`.

src/js/reducers/scalesReducer.ts
```diff
@@ -31,7 +31,13 @@ export function scalesReducer(state: ScaleState, action: ActionType<typeof scale
...
   if (action.type === getType(scaleActions.updateScaleProperty)) {
     const p = action.payload;
-    return state.setIn([String(id), p.property], p.value);
+    if (state.getIn([String(id), ...p.property.split(".")])) {
+      return state.setIn([String(id), ...p.property.split(".")], p.value);
+    }
+    const key = p.property.split(".").slice(-1).toString();
+    const obj = {};
+    obj[key] = p.value;
+    return state.setIn([String(id), ...p.property.split(".").slice(0,-1)], obj);
```
jonathanzong: these lines break things right now. to remind me, can you write down the cases that you're trying to handle here?

ktbacher: Yeah this is what I needed some help with... Trying to handle the `range = {scheme: someColorScheme}` vs `range = someString`

jonathanzong: cool -- i think you can change this from an if statement to a try catch to handle the `TypeError`:
```
try {
      return state.setIn([String(id), ...p.property.split(".")], p.value);
    }
    catch (e) {
      const key = p.property.split(".").slice(-1).toString();
      const obj = {};
      obj[key] = p.value;
      return state.setIn([String(id), ...p.property.split(".").slice(0,-1)], obj);
    }
```

jonathanzong: and maybe leave a comment quickly explaining the typescript issue and reference the range and range.scheme example

NanaKwame:
In my recent commit:
- added keypress event support to FormInput, Property and MultiValueProperty.
- added onblur support to Property, MultiValueProperty
- added code export.ts to export the right vega specs for manual vs data field domain

NanaKwame:
@jonathanzong  Currently the questions I have are:
- Do I need to fix MultiValueProperty.test.jsx? If so how do I run this to make sure it is doing the right thing?
- I went a bit overboard with the event support but I think both onBlur and onKeyPress are valuable. lmk?

@ktbacher do we still need the value.scheme fix. I am getting an error because value is declared as a const. I might also have missed some updates with my merge conflict resolution.

src/js/components/inspectors/MultiValueProperty.test.jsx
```diff
@@ -0,0 +1,239 @@
+/* eslint no-unused-expressions:0,new-cap:0 */
```
NanaKwame: not sure if I should get these tests to work or how do I run them?

src/js/store/factory/Scale.ts
```diff
@@ -16,7 +16,9 @@ export type LyraScale = {
    */
   _origName: string;
   _domain: any[],
-  _range: any[]
+  _range: any[],
+  _manual: boolean,
+  _domainArray: any[]
```
NanaKwame: introduced this so we don't need to override _domain is the user switches to manual. That way if they toggle the manual check the _domain value with fields isn't lost. But we resolve this in the exports.ts file.

src/js/components/inspectors/MultiValueProperty.test.jsx
jonathanzong: the tests are stale so ignore them!

src/js/store/factory/Scale.ts
jonathanzong: is `_manualDomain` a clearer variable name for `_domainArray`?

src/js/components/inspectors/MultiValueProperty.tsx
```diff
@@ -0,0 +1,178 @@
...
+    if (type === 'number' || type === 'range') {
+      value = +value;
+    }
+
+    let newVal = props.processValue ? props.processValue(value, props) : value;
+
+    let arrayValues = [...props.value];
+    arrayValues.push(newVal);
+
+    evt.target.name = props.name;
```
NanaKwame, line 76: working on a better way to go about this.

src/js/components/inspectors/FormInputProperty.tsx
```diff
@@ -21,6 +21,7 @@ interface OwnProps {
   signal?: string;
   onChange?: (e) => any; // TODO: find function in/out types
   onBlur?: () => any; // TODO: find function in/out types
+  onKeyPress?: () => any; // TODO: find function in/out types
```
jonathanzong: these should all be `=> void` if you wanna resolve the TODOs :)

src/js/components/inspectors/Property.tsx
```diff
@@ -71,7 +73,7 @@ function mapStateToProps(reduxState: State, ownProps: OwnProps): StateProps {
   const field = getIn(state, path + '.field');
   const scaleName = scale && getInVis(reduxState, 'scales.' + scale + '.name');
 
-  const value = getIn(state, path);
+  const value = ownProps.value === undefined ? getIn(state, path) : null;
```
jonathanzong: if you read this and then read line 79, this change appears to amount to this effect: return ownProps.value if it's not undefined, else get it from the store. if the store value is null, return undefined. If that's what we want, we should try rewriting it as a less complex expression

src/js/components/inspectors/Scale.tsx
```diff
@@ -1,46 +1,173 @@
...
+import { MultiValueProperty } from './MultiValueProperty';
 import {ScaleRecord} from '../../store/factory/Scale';
+import { updateScaleProperty } from '../../actions/scaleActions';
+import { ScaleSimpleType, scaleTypeSimple} from '../../store/factory/Scale';
 
 interface OwnProps {
   primId: number;
   primType: PrimType;
-
+  _manual: boolean;
```
jonathanzong: no references to this -- remove?

src/js/components/inspectors/Scale.tsx
```diff
@@ -1,46 +1,173 @@
...
 }
 
 interface StateProps {
   scale: ScaleRecord;
+  fields: any;
+  fieldMap: any;
```
jonathanzong: fields: string[];
fieldMap: Map<string, string[]>;

src/js/components/inspectors/MultiValueProperty.tsx
```diff
@@ -0,0 +1,178 @@
...
+      (props.firstChild ? ' first-child' : '');
+
+    return (
+      <div>
+        {labelEl}
+        <div className={className}
+          onDragEnter={() => this.setState({ dragOver: ++dragOver })}
+          onDragLeave={() => this.setState({ dragOver: --dragOver })}>
+          {selectedValsEl}
+          <Property {...props} label="Add" value={null} onChange={props.onChange ? (e) => this.handleChange(e) : null} onBlur={props.onBlur ? (e) => this.handleChange(e) : null} onKeyPress={props.onKeyPress ? (e) => this.handleChange(e) : null} />
```
jonathanzong: why `onKeyPress`? did the existing `onChange` handling in `FormInputProperty` not work?

src/js/components/inspectors/MultiValueProperty.test.jsx
```diff
@@ -0,0 +1,239 @@
+// /* eslint no-unused-expressions:0,new-cap:0 */
```
jonathanzong, line 1: i just realized that i misunderstood your question earlier. i thought you were asking about old tests, but i see that you wrote new tests!

i've never run the tests before but i think that you can try to run these using `yarn test`. see `package.json` for a list of scripts that you can run with yarn

NanaKwame: so i didn't actually write this. i just copied them from property and did a string search replace so they are actually wrong i think. :/ so will probably leave it commented out.

NanaKwame: wdyt?

src/js/components/inspectors/MultiValueProperty.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { Property } from './Property';
+
+interface OwnState {
+  dragOver?: number;
+}
+
+// hunch: probably actually needs to be a bunch of different types anded to a base type
+// instead of one type with everything optional
+// but it would take a long time to figure out what the mutually exclusive fields are
+interface OwnProps {
```
jonathanzong, line 11: this props interface is a mess because `Property.tsx` doesn't have a consistent interface across all its uses. Since we're writing this new component, and can control where it's used, let's clean up the props interface to remove optionals and remove props that aren't used here.

src/js/components/inspectors/MultiValueProperty.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+interface StateProps {
+  group?: any;
+  signal?: any;
+  value?: string | number | boolean | any; // TODO: remove 'any', add Immutable.Map type
+  field?: any;
+  scale?: any;
+  srcField?: any;
+  scaleName?: any;
+  dragging?: DraggingStateRecord;
```
jonathanzong: For example, do you want people to be able to drag fields or signals onto the `MultiValueProperty`? If not, let's not leave this prop here.

src/js/components/inspectors/MultiValueProperty.tsx
NanaKwame: so the issue with onchange is that it fires for every change(even while you are typing) which we don't want for the manual text input. so just adding support for keypress so that we can listen for the enter key.

src/js/components/inspectors/MultiValueProperty.tsx
NanaKwame: Gotcha. So I am going to take this as strip it down to just what we need for now and we can add more stuff later if we need it? Gonna go with that for now.

src/js/components/inspectors/Property.tsx
NanaKwame: oh yeah i missed that. thnx!

src/js/components/inspectors/MultiValueProperty.tsx
NanaKwame: I removed a bunch of stuff. Lmk if I should remove more or put some stuff back.

NanaKwame:
known console errors:
- `value` prop on `input` should not be null...
- A component is changing a controlled input of type checkbox to be uncontrolled...

looking into these but I think we can merge with these. could be separate issues.

src/js/components/inspectors/MultiValueProperty.test.jsx
jonathanzong: ah ok, that's fine to just leave it then! all the tests haven't been updated since before the refactor, and since we're actively making changes at a fast pace, it's not worth investing much time into updating the tests until things are a bit more stable IMO

src/js/components/inspectors/Property.tsx
```diff
@@ -71,12 +73,10 @@ function mapStateToProps(reduxState: State, ownProps: OwnProps): StateProps {
   const field = getIn(state, path + '.field');
   const scaleName = scale && getInVis(reduxState, 'scales.' + scale + '.name');
 
-  const value = getIn(state, path);
-
   return {
     group:  getIn(state, path + '.group'),
     signal: getIn(state, path + '.signal'),
-    value:  value !== null && value !== undefined ? value : ownProps.value,
+    value: ownProps.value === undefined ? getIn(state, path) : undefined, // Not using || because value could be 0
```
jonathanzong: observation: it's now no longer possible for the value of `ownProps.value` to be directly returned if it's not `undefined`. is that what we want?

NanaKwame: thnx for putting in the fix for this.

PR #528, opened 2020-11-24
Delete Scale UI

### LINKED ISSUE #504: UI to Delete Scales
Want to add a UI component so users can delete scales

NanaKwame:
Adding UI for deleting scales.

### DIFF
--- a/src/js/components/encodings/ScaleList.tsx
+++ b/src/js/components/encodings/ScaleList.tsx
@@ -8,14 +8,21 @@ import sg from '../../ctrl/signals';
 import {channelName} from '../../actions/bindChannel';
 import {MODE, SELECTED, CELL} from '../../store/factory/Signal';
 import {ScaleDraggingState, DraggingStateRecord, ScaleDraggingStateRecord} from '../../store/factory/Inspector';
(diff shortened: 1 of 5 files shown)


NanaKwame:
FYI in guidesReducer.ts, guideReducer deletes guides when scales are deleted.

src/js/components/encodings/ScaleList.tsx
```diff
@@ -109,7 +143,12 @@ class ScaleList extends React.Component<StateProps & DispatchProps> {
                   data-scale={scale.get('_id')}
                   data-field={field}>
                   {/* <ContentEditable value={name} save={updateScaleName}  /> */}
-                  {name}
+                  <div style={{ "marginLeft": "26px" }}>
```
jonathanzong, line 146: let's try to avoid inline styles if possible. can we put this in the stylesheet?

jonathanzong:
this looks good! requested behaviors:

- when deleting a color scale, the marks disappear. if the encoding is "fill" can we replace the value with a default instead of dropping the encoding?
- after deleting a size scale, manually changing the size using the inspector no longer works

PR #532, opened 2021-03-10
Drag & Drop Marks

ktbacher:
Adds a component for a mark dropzone and makes marks draggable from the toolbar.

### DIFF
--- a/src/js/components/toolbar/AddMarks.tsx
+++ b/src/js/components/toolbar/AddMarks.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import {addMark} from '../../actions/markActions';
 import { LyraMarkType, Mark } from '../../store/factory/Mark';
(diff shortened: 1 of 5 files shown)


src/js/components/toolbar/AddMarks.tsx
```diff
@@ -27,21 +31,42 @@ function mapDispatchToProps(dispatch, ownProps): DispatchProps {
...
+    this.props.addMark(mark);
+
+  }
+  public handleDragStart = (mark) => {
+    this.props.startDragging(MarkDraggingState({mark}));
+  }
+
+  // This makes use of the bubble cursor, which corresponds to the cell signal;
+  // we're using that to figure out which channel we are closest to. The
+  // SELECTED signal indicates the mark to bind the data to.
```
jonathanzong: these (i think copied/pasted?) comments don't apply to what you're doing, remove to reduce confusion?

jonathanzong (approved):

PR #533, opened 2021-03-12
Layout store, actions, and reducers

ktbacher:
Provides the base scaffolding for the layout mechanism, including:
- Layout store element to keep track of key aspects 
- Layout actions and reducers to handle adding, positioning, etc
- Dropzones for adding next groups to a layout and positioning the accordinly


src/js/actions/layoutActions.ts
```diff
@@ -0,0 +1,67 @@
...
+export function addLayout (payload: LayoutRecord) {
+  return function(dispatch: Dispatch, getState: () => State) {
+    const id = payload._id || assignId(dispatch, getState());
+    dispatch(baseAddLayout(payload.merge({_id: id}), id));
+  };
+}
+export const baseAddLayout = createStandardAction('ADD_LAYOUT')<LayoutRecord, number>();
+
+
+export function addGrouptoLayout (payload: {group: MarkRecord, dir: string}, id: number) {
```
jonathanzong: what if we use the more specific type annotation `GroupRecord` instead of `MarkRecord`?

src/js/actions/markActions.ts
```diff
@@ -36,6 +37,18 @@ export function addMark (record: MarkRecord) {
     }, id));
   };
 }
+
+export function addGroup(record: MarkRecord, layoutId: number, dir: string) {
+  return function(dispatch: Dispatch, getState: () => State) {
```
jonathanzong: i remember now: 
- `import {ThunkDispatch} from 'redux-thunk';`
- change the type annotation for `dispatch` from `Dispatch` to `ThunkDispatch<State, any, any>`
- remove the two `as any` below

src/js/index.ts
```diff
@@ -31,6 +33,15 @@ store.subscribe(listeners.createStoreListener(store, ctrl));
...
 (store as any).dispatch(addMark(Mark('group', {_parent: 1})));
+var state = store.getState();
+var group = state.getIn(['vis', 'present', 'marks', '2']);
+(store as any).dispatch(addLayout(Layout({
+  rows: 1,
+  cols: 1,
+  groups: [2],
+  rowSizes: [group.encode.update.height],
+  colSizes: [group.encode.update.width]
+}))); // fix hardcoded group ID
```
jonathanzong: try this?

```
const layoutId = assignId(store.dispatch, store.getState())
(store as any).dispatch(addLayout(Layout({
  _id: layoutId,
  rows: 1,
  cols: 1,
})));
(store as any).dispatch(addGroup(Mark('group', {_parent: 1}), layoutId, null));
```

letting `addGroup` handle adding the group to the layout

src/js/reducers/layoutReducer.ts
```diff
@@ -0,0 +1,64 @@
+import {Map} from 'immutable';
```
jonathanzong: for consistency with the rest of the reducer filenames, i would name this `layoutsReducer.ts` (layouts plural). right-click rename in vscode should prompt you to automatically update all imports. do this in a fresh commit so you don't mix up the auto-rename with other changes in case something goes wrong

src/js/reducers/layoutReducer.ts
```diff
@@ -0,0 +1,64 @@
...
+import {ActionType, getType} from 'typesafe-actions';
+import {LayoutState} from '../store/factory/Layout';
+import * as layoutActions from '../actions/layoutActions';
+
+/**
+ * This reducer handles layout updates
+ * @param {Object} state - An Immutable state object
+ * @param {Object} action - An action object
+ */
+export function layoutReducer(state: LayoutState,
```
jonathanzong: similarly, i would name this function `layoutsReducer`

src/js/reducers/layoutReducer.ts
```diff
@@ -0,0 +1,64 @@
...
+  action: ActionType<typeof layoutActions>): LayoutState {
+  // handle layout actions
+  const id = action.meta;
```
jonathanzong: consider doing `const id = String(action.meta);` since you cast it to string everywhere below

src/js/reducers/layoutReducer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (typeof state === 'undefined') {
+    return Map();
+  }
+
+  if (action.type === getType(layoutActions.baseAddLayout)) {
+    return state.set(String(id), action.payload);
+  }
+  else if (action.type === getType(layoutActions.baseAddGrouptoLayout)) {
```
jonathanzong: this is not a big deal, but stylistically and for consistency purposes i prefer just using separate if blocks instead of chaining `else if` here because there's return statements at the bottom of each block. conceptually, reducers are more like switch-cases over the action types. it also helps you more easily skim the difference between `if`s that check the action types and if-else logic within each action handler

src/js/store/factory/Layout.ts
```diff
@@ -0,0 +1,73 @@
...
+   * Size and positioning of this placeholder.
+   */
+  top: number;
+  left: number;
+  width: number;
+  height: number;
+
+}
+
+export const placeHolder = Record<LyraPlaceholder>({
```
jonathanzong: consistent choice of `placeHolder` vs `Placeholder` vs `PlaceHolder`? fwiw i prefer `Placeholder`

in case you don't know about it, right-click and "rename symbol" in vscode will also rename all the references for you

src/js/components/toolbar/MarkDropPlace.tsx
```diff
@@ -0,0 +1,120 @@
...
+import {State} from '../../store';
+import {MarkDraggingStateRecord} from '../../store/factory/Inspector';
+import { LyraMarkType, Mark } from '../../store/factory/Mark';
+import {addMark, addGroup } from '../../actions/markActions';
+import {removePlaceHolder} from '../../actions/layoutActions';
+import {getClosestGroupId} from '../../util/hierarchy';
+import { LayoutRecord} from '../../store/factory/Layout';
+
+const imutils = require('../../util/immutable-utils');
+const getInVis = imutils.getInVis;
```
jonathanzong: we're trying to phase out `getInVis(...)` in favor of `state.getIn(['vis', 'present', ... ])` going forward, because the latter plays more nicely with type checking

src/js/components/toolbar/MarkDropPlace.tsx
```diff
@@ -0,0 +1,120 @@
...
+      evt.preventDefault();
+    }
+
+    return false;
+  };
+
+  public handleDrop = ()  => {
+    const sceneId = this.props.sceneId;
+    this.props.addGroup(sceneId);
+    this.props.addMark(this.props.dragging.mark, null);
```
jonathanzong, line 105: can just pass in id `this.props.addMark(this.props.dragging.mark, sceneId);`

src/js/components/toolbar/MarkDropPlace.tsx
```diff
@@ -0,0 +1,120 @@
...
+  };
+}
+
+function mapDispatchToProps(dispatch, ownProps: OwnProps): DispatchProps {
+  return {
+    addMark: (type, parentId) => {
+;      if (!parentId) {
+        parentId = getClosestGroupId();
+      }
+
```
jonathanzong, line 52: don't really need this if you pass in parentId below

src/js/components/toolbar/MarkDropPlaceGroup.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+function mapStateToProps(state: State): StateProps {
+  const layouts = state.getIn(['vis', 'present', 'layouts']);
+
+  const layoutId = Array.from(layouts.keys())[0];
+  // const layout = layouts.filter(layout => layout._id == ind);
+  const placeholders = state.getIn(['vis', 'present', 'layouts', layoutId, 'placeHolders']);
+
+  return {
+    layoutId: +layoutId,
```
jonathanzong, line 20: maybe `Number(layoutId)` instead of `+layoutId`?

jonathanzong: `Number(layoutId)` instead of `+layoutId`?

i just realized that everywhere in lyra, we type `id`s as numbers in actions/components but strings in the store. rip. not something you have to worry about but i will now not be able to unsee that

src/js/store/factory/Layout.ts
```diff
@@ -0,0 +1,73 @@
...
+  _id: null,
+  top: 0,
+  left: 0,
+  width: 200,
+  height: 150
```
jonathanzong: are these default values going to need to stay synchronized with the default group width/height in `Group.ts`? if so, consider making `export const`s for width/height in `Group.ts`

jonathanzong:
great work! my comments are pretty minor, so happy to do another pass later or for you to just address them and merge as you see fit

src/js/components/toolbar/MarkDropPlace.tsx
```diff
@@ -0,0 +1,120 @@
...
+  addMark: (type: LyraMarkType, parentId: number) => void;
+  addGroup: (sceneId: number) => void;
+  removePlaceHolder: () => void;
+}
+
+function mapStateToProps(state: State,  ownProps: OwnProps): StateProps {
+  const draggingRecord = state.getIn(['inspector', 'dragging']);
+  const isMarkDrag = draggingRecord && (draggingRecord as MarkDraggingStateRecord).mark;
+  let sceneId;
+  state.getIn(['vis', 'present', 'marks']).filter(mark => mark.name == "Scene").forEach(scene => sceneId = scene._id);
```
ktbacher: This was my work around instead of using getInVis but there is probably a cleaner and clearer way to do this?

jonathanzong: `state.getIn(['vis', 'present', 'scene', '_id'])`

PR #535, opened 2021-03-30
Esefah meng history strip

NanaKwame:
This PR includes:
- changes for showing the history at the bottom of the page
- DnD of the history items.
- Being able to drag history items to x, y, fill or size on the bubble cursor and see scale updates
- Being able to drag history items to the main chart but not on the bubble cursor to see aesthetic updates to marks: "size", "shape", "fill", "fillOpacity", "stroke",  "strokeWidth"

TODOs amongst bigger todos:
- fix any issues that might come up with changes to exporter.ts and cleanSpecForPreview in demonstrations.ts
- figure out if I need to use History.ts
- other mark types aren't working perfectly. currently symbols work the best


src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+import {setSignal} from '../../actions/signalActions';
+import {SignalValue} from 'vega-typings/types';
+import {batchGroupBy} from '../../reducers/historyOptions';
+
+const ctrl = require('../../ctrl');
+
+
+interface OwnProps {
+  id: string,
+  history: HistoryRecord // TODO: use History.ts for you just have to pass the id
```
jonathanzong: It looks like this prop doesn't currently use a HistoryRecord. Maybe remove the type annotation for now so that it's not confusing (to a reader or to the type checker)?

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  groupNames: any[];
```
jonathanzong: `string[]`

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+      dispatch(setMarkVisual(p, markId));
+    },
+    bindChannel: (dsId: number, field: ColumnRecord, markId: number, property: string) => {
+      dispatch(bindChannel(dsId, field, markId, property));
+    },
+    setSignal: (value: SignalValue, signal: string) => {
+      dispatch(setSignal(value, signal));
+    }
+  };
+}
```
jonathanzong: (style)
Since your dispatch functions have no side effects, you could also use the `actionCreators` syntax here. See `Interaction.tsx` line 139 for an example.

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+export class HistoryItemInspector extends React.Component<OwnProps & DispatchProps> {
+
+  constructor(props) {
+    super(props);
+  }
+
+  private width = 100; // these should match baseSignals in demonstrations.ts
```
jonathanzong: does this comment apply here? if not, remove it

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+  private height = 100; //
+
+  public handleClick = (historyId: number) => {
+
+  }
+  public handleDragStart = (historyId: number) => {
+    this.props.startDragging(HistoryDraggingState({historyId}));
```
jonathanzong: this looks unused

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+        const bindField = this.props.history.getIn(["datasets", String(dsId), "_schema", fieldName]).toJS();
+        vega.extend(bindField, opts); // Aggregate or Bin passed in opts.
+        props.bindChannel(dsId, bindField, lyraId, cell.key);
+      } else {
+        // update the whole symbol/mark: shape, size
+        // the history has the signals
+        let relevantProps = ["size", "shape", "fill", "fillOpacity", "stroke",  "strokeWidth"];
+        let historyMark = this.props.history.getIn(["marks", String(lyraId)]); // assume mark is the same as before. wanna revert to old settings of old mark
+
+        batchGroupBy.start();
```
jonathanzong: this is unnecessary because `setSignal` is already batched

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+        relevantProps.forEach((prop) => {
+          let historyProp = historyMark.getIn(["encode", "update", prop]);
+          if (historyProp.signal) {
+            let historySigVal = this.props.history.getIn(["signals", historyProp.signal]).value;
+            this.props.setSignal(historySigVal, historyProp.signal); // update Store
+            sg.set(historyProp.signal, historySigVal, false); // update Vega
+          }
+        });
+        batchGroupBy.end();
```
jonathanzong: see above

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+        let relevantProps = ["size", "shape", "fill", "fillOpacity", "stroke",  "strokeWidth"];
+        let historyMark = this.props.history.getIn(["marks", String(lyraId)]); // assume mark is the same as before. wanna revert to old settings of old mark
+
+        batchGroupBy.start();
+        relevantProps.forEach((prop) => {
+          let historyProp = historyMark.getIn(["encode", "update", prop]);
+          if (historyProp.signal) {
+            let historySigVal = this.props.history.getIn(["signals", historyProp.signal]).value;
+            this.props.setSignal(historySigVal, historyProp.signal); // update Store
+            sg.set(historyProp.signal, historySigVal, false); // update Vega
```
jonathanzong: i suggest doing
```
sg.set(historyProp.signal, historySigVal, false);
ctrl.update();
```
instead of dispatching `setSignal`

jonathanzong: unless there's something that doesn't work without `setSignal`. i'm suggesting this because this is what the inspectors do to update these types of values

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+            return scaleRecord.get('_domain').length > 0 ?  scaleRecord.get('_domain')[0].field : null;
+          }).first();
+
+          dsId = channelScaleIds.map((scaleId) => {
+            let scaleRecord = this.props.history.getIn(["scales"]).get(scaleId);
+            return scaleRecord.get('_domain').length > 0 ?  scaleRecord.get('_domain')[0].data : null;
+          }).first();
+
+
+        }
```
jonathanzong: this is fine for now as you're prototyping, but from a design perspective I don't think we should use dropping outside of a dropzone as the final interaction for non-x/y channels

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const cell = sg.get(CELL);
+    const dropped = tupleid(sel) && tupleid(cell);
+
+    try {
+      const lyraId = +sel.mark.role.split('lyra_')[1]; // id of thing that was dropped onto
+
+      if (dropped) {
+        const channel = channelName(cell.key);
+        let fieldName, dsId;
+        if (channel === 'x' || channel === 'y' || channel === 'color' || channel === 'size') {
```
jonathanzong: for other mark types, the channels won't always be `color` and `size`. i think if you're doing an if-case you should only case on `x` and `y`, and handle drops on the dropzones that are over the mark itself in the else case.

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+          }
+        });
+        batchGroupBy.end();
+      }
+    } catch (e) {
+      console.error('Unable to bind primitive');
+      console.error(e);
+    }
+
+    this.props.stopDragging();
```
jonathanzong, line 114: dragging state record is unused

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+    sg.set(MODE, 'handles');
+    sg.set(CELL, {});
+
+    ctrl.update(); // Apply changes
+
+  }
+
+  private historyToSpec(preview: HistoryRecord): Spec {
```
jonathanzong: parameter is unused

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+import {State} from '../../store';
+import {setSignal} from '../../actions/signalActions';
+import {SignalValue} from 'vega-typings/types';
+import {batchGroupBy} from '../../reducers/historyOptions';
+
+const ctrl = require('../../ctrl');
+
+
+interface OwnProps {
+  id: string,
```
jonathanzong: consider passing `id` as a `number` if you call parseInt on it in all places where it's used

src/js/components/history/HistoryItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+    sg.set(CELL, {});
+
+    ctrl.update(); // Apply changes
+
+  }
+
+  private historyToSpec(preview: HistoryRecord): Spec {
+    let historySpec = ctrl.export(false, this.props.history);
+    if (historySpec.marks) {
+      historySpec = cleanSpecForPreview(historySpec, this.width, this.height, null, parseInt(this.props.id), true);
```
jonathanzong: see the comment near the top of the file about whether `id` should be typed as a `number`

src/js/components/history/HistoryList.tsx
```diff
@@ -0,0 +1,56 @@
...
+interface StateProps {
+  history: any[];
+  groupNames: string[];
+}
+
+interface DispatchProps {
+  updateHistoryProperty: (payload: {property: string, value: any}, id: number) => void;
+}
+
+function mapState(state: State): StateProps {
```
jonathanzong: `mapStateToProps` just for consistency across the codebase

src/js/components/history/HistoryList.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }).toList().toJSON();
+  let history = [...getIn(state, 'vis').past];
+  history.push(getIn(state, 'vis').present);
+  return {
+    history,
+    groupNames: groupNames
+  };
+}
+
+const mapDispatch: DispatchProps = {
```
jonathanzong: for consistency across the codebase, you could name this `actionCreators`

src/js/store/factory/Inspector.ts
```diff
@@ -46,15 +50,20 @@ export const SignalDraggingState = Record<LyraSignalDraggingState>({
   signal: null
 }, 'LyraSignalDraggingState');
 
+export const HistoryDraggingState = Record<LyraHistoryDraggingState>({
```
jonathanzong, line 53: i recommend not introducing new types into the codebase until it's clear you need to use them. `HistoryDraggingState` is currently not used anywhere

src/js/components/history/HistoryList.tsx
```diff
@@ -0,0 +1,56 @@
...
+interface DispatchProps {
+  updateHistoryProperty: (payload: {property: string, value: any}, id: number) => void;
+}
+
+function mapState(state: State): StateProps {
+  const marks: Map<string, MarkRecord> = state.getIn(['vis', 'present', 'marks']);
+  const groupNames = marks.filter((mark: MarkRecord) => {
+    return mark.type === 'group';
+  }).map((v) => {
+    return v.name.replace(" ", "_");
```
jonathanzong, line 27: i think these names should already be human readable. also, users can rename marks. double check whether this replace might not be necessary

jonathanzong: `exportName()` from `util/exportName.ts`

src/js/components/history/HistoryList.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+class BaseHistoryList extends React.Component<OwnProps & StateProps & DispatchProps> {
+  public render() {
+
+    return (
+      <div id='history-list' >
+        {this.props.history.map(
+           (item, idx) => {
+            return <HistoryItem id={idx+''} key={idx+''} history={item} groupNames={this.props.groupNames} />
```
jonathanzong: see other comments above about the type of `id`

src/js/components/HistoryToolbar.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import {HistoryList} from './history/HistoryList';
+
+export class HistoryToolbar extends React.PureComponent<{}, {modalIsOpen: boolean}> {
```
jonathanzong: this component only wraps another component. merge into one `HistoryList` component?

src/js/ctrl/demonstrations.ts
```diff
@@ -1211,7 +1211,7 @@ export function getScaleInfoForGroup(state: State, groupId: number): ScaleInfo {
   };
 }
 
-export function cleanSpecForPreview(sceneSpec, width: number, height: number, groupName: string, interactionId: number): Spec {
+export function cleanSpecForPreview(sceneSpec, width: number, height: number, groupName: string, interactionId: number, isHistoryItem: boolean): Spec {
```
jonathanzong: i think you should actually have a separate function for cleaning history specs instead of overloading this function in `demonstrations`. it will make more conceptual sense, limit the amount of edits you will have to make to files that do things unrelated to history and limit the risks of introducing bugs unrelated to history when you make changes

src/js/components/interactions/InteractionPreview.tsx
```diff
@@ -26,7 +26,7 @@ export class InteractionPreview extends React.Component<OwnProps> {
 
   private previewToSpec(preview: SelectionRecord | ApplicationRecord): Spec {
     const groupName = (preview as TransformApplicationRecord).targetGroupName || this.props.groupName;
-    let spec = cleanSpecForPreview(ctrl.export(false), this.width, this.height, groupName, this.props.interaction.id);
+    let spec = cleanSpecForPreview(ctrl.export(false), this.width, this.height, groupName, this.props.interaction.id, false);
```
jonathanzong: see below comment on `cleanSpecForPreview`

src/js/components/interactions/WidgetPreview.tsx
```diff
@@ -21,7 +21,7 @@ export class WidgetPreview extends React.Component<OwnProps> {
 
   private previewToSpec(preview: MarkApplicationRecord): Spec {
     const groupName = this.props.groupName;
-    let spec = cleanSpecForPreview(ctrl.export(false), this.width, this.height, groupName, this.props.widget.id);
+    let spec = cleanSpecForPreview(ctrl.export(false), this.width, this.height, groupName, this.props.widget.id, false);
```
jonathanzong: see below comment on `cleanSpecForPreview`

src/js/ctrl/demonstrations.ts
```diff
@@ -1296,20 +1325,27 @@ export function cleanSpecForPreview(sceneSpec, width: number, height: number, gr
     return markSpec;
   });
 
-  return addBaseSignalsForPreview(sceneUpdated, groupName, interactionId);
+  return addBaseSignalsForPreview(sceneUpdated, groupName, interactionId, isHistoryItem);
```
jonathanzong: same idea as `cleanSpecForPreview`. you don't need to be adding base interaction signals to non-interactive history previews

src/js/ctrl/demonstrations.ts
```diff
@@ -1296,20 +1325,27 @@ export function cleanSpecForPreview(sceneSpec, width: number, height: number, gr
...
 }
 
-function addBaseSignalsForPreview(sceneSpec, groupName, interactionId) {
+function addBaseSignalsForPreview(sceneSpec, groupName, interactionId, isHistoryItem) {
```
jonathanzong: same as above

src/js/ctrl/export.ts
```diff
@@ -47,19 +51,20 @@ export function exporter(internal: boolean = false): Spec {
   spec.data = exporter.pipelines(state, int);
 
   // Add interactions and widgets from store
-  spec = exporter.interactions(state, spec);
+  spec = history ? spec : exporter.interactions(state, spec);
```
jonathanzong: why make an exception for interactions? when someone reverts to a prior history before they added interactivity, won't they want that interactivity to be gone?

src/js/ctrl/export.ts
```diff
@@ -47,19 +51,20 @@ export function exporter(internal: boolean = false): Spec {
...
   spec = exporter.widgets(state, spec);
   return spec;
 }
 
 exporter.interactions = function(state: State, spec) {
-  state.getIn(['vis', 'present', 'interactions']).forEach((interaction: InteractionRecord) => {
-    const group: GroupRecord = state.getIn(['vis', 'present', 'marks', String(interaction.groupId)]);
+  let storeState = store.getState(); // use Store state instead
```
jonathanzong, line 60: same comment as above

src/js/ctrl/export.ts
```diff
@@ -105,7 +110,7 @@ exporter.widgets = function(state: State, spec) {
 }
 
 exporter.pipelines = function(state: State, internal: boolean) {
-  const pipelines: PipelineRecord[] = getInVis(state, 'pipelines').valueSeq();
+  const pipelines: PipelineRecord[] = state.getIn(['pipelines']).valueSeq();
```
jonathanzong: this is the right idea. however, you've changed the type of `state` from `State` to `VisStateTree` and should update type annotations to match (this applies elsewhere in the file too)

src/js/ctrl/export.ts
```diff
@@ -37,8 +37,12 @@ let counts = duplicate(SPEC_COUNT);
...
  * @returns {Object} A Vega specification.
  */
-export function exporter(internal: boolean = false): Spec {
-  const state = store.getState();
+export function exporter(internal: boolean = false, history): Spec { // TODO: use history.ts so you only have to pass id or index
+  if (history && history.getIn(['marks']).size == 0) {
+    return {};
+  }
+
+  const state = history || store.getState().getIn(['vis', 'present']);
```
jonathanzong: i would change the parameter from `history` to `visState?: VisStateTree` and do `visState = visState || store.getState().getIn(['vis', 'present']);`

this makes it clearer what the export function does.

src/js/store/factory/History.ts
```diff
@@ -0,0 +1,25 @@
...
+   */
+  _id: number;
+  /**
+   * The index of the history in the maintained history list.
+   */
+  index: number;
+
+}
+
+export const History = Record<LyraHistory>({
```
jonathanzong, line 18: i think figuring out what this record does will be an important next step

jonathanzong (changes requested):
nice work! i've left comments that i hope are helpful. since it's a working branch, i'll leave it up to you what level of detail you want to address them to before merging

src/js/ctrl/export.ts
NanaKwame: so this was causing the thumbnails to be interactive too. So when you try to drag and drop the selection marquee activates making it impossible to drag and drop.

src/js/ctrl/export.ts
NanaKwame: got it. yeah wasn't sure what the new state was. sg

src/js/components/history/HistoryItem.tsx
NanaKwame: yeah it still applies but might change after some of the other refactoring you suggested.

src/js/components/history/HistoryItem.tsx
NanaKwame: Yeah I wasn't exactly sure what it was used for. Also if I go down using custom drop zones I would need this right? Some of the stuff in this PR is for just in case I need them down the line.

src/js/components/history/HistoryItem.tsx
NanaKwame: oh I thought they used both. I will recheck.

src/js/components/history/HistoryItem.tsx
NanaKwame: +1

src/js/components/history/HistoryItem.tsx
NanaKwame: yeah currently this is kinda aimed at symbols but will adapt for other mark types in PRs to follow

src/js/store/factory/Inspector.ts
NanaKwame: gotcha. sg

src/js/components/HistoryToolbar.tsx
NanaKwame: +1

src/js/ctrl/demonstrations.ts
NanaKwame: would this be in another file I create in src/js/ctrl?

src/js/ctrl/export.ts
jonathanzong: ah, thanks for clarifying! i think that's something you'll want to take care of in your version of `cleanSpecForPreview`, by removing signals from the previews. doing this in the exporter will have downstream effects for other things, like traversing the history

src/js/ctrl/demonstrations.ts
jonathanzong: i think it can probably just go in one of your components for now. i moved the interactions one to ctrl so that i could share it between interactions and widgets

src/js/components/history/HistoryItem.tsx
NanaKwame: leaving this for when I do custom dropzones later

src/js/components/history/HistoryItem.tsx
NanaKwame: leaving this for when I implement a tree structure for history

src/js/components/history/HistoryItem.tsx
NanaKwame: didn't need setSignal!

src/js/components/history/HistoryItem.tsx
NanaKwame: leaving this for when I implement custom drop zones later.

src/js/store/factory/Inspector.ts
NanaKwame: going to keep this for when I create custom dropzones later

src/js/ctrl/export.ts
NanaKwame: not sure exactly what this reference but I think I resolve it in my next PR. Lmk if this is still an issue.

NanaKwame:
> nice work! i've left comments that i hope are helpful. since it's a working branch, i'll leave it up to you what level of detail you want to address them to before merging

I have pushed changes to address your comments. Please send any feedback especially for cleanHistorySpec in HistortyList.tsx. Not sure if that is the right/ok way to go about removing interactions from the history previews.

src/js/ctrl/export.ts
```diff
@@ -37,29 +37,34 @@ let counts = duplicate(SPEC_COUNT);
...
-    const group: GroupRecord = state.getIn(['vis', 'present', 'marks', String(interaction.groupId)]);
+exporter.interactions = function(visState: VisStateTree, spec) {
+  let storeState = store.getState(); // use Store state instead
+  visState.getIn(['interactions']).forEach((interaction: InteractionRecord) => {
+    const group: GroupRecord = visState.getIn(['marks', String(interaction.groupId)]);
     const groupName = exportName(group.name);
-    const scaleInfo = getScaleInfoForGroup(state, group._id);
-    const fieldsOfGroup = getFieldsOfGroup(state, group._id);
+    const scaleInfo = getScaleInfoForGroup(storeState, group._id);
+    const fieldsOfGroup = getFieldsOfGroup(storeState, group._id);
```
jonathanzong, line 65: now that i think about it, these should use visState too. if we pass in an older visState to the exporter, we expect it to get the scale info for that state. using the store state will use the current scale info instead, which could cause problems.

this means that the getScaleInfoForGroup and getFieldsOfGroup functions should probably be edited to use visState as well, and all uses of those functions updated accordingly. could you add one more commit with that change?

jonathanzong:
great work implementing the earlier feedback! your approach in cleanHistorySpec looks right to me. left one comment for one last refactoring task then you're good to go

PR #538, opened 2021-04-16
Data faceting

ktbacher:


src/js/components/pipelines/FacetDropzone.tsx
```diff
@@ -0,0 +1,76 @@
...
+    facetField: (field, groupId) => {
+      let numCols;
+      if (ownProps.layoutOrientation == "Column") {
+        numCols = 1;
+      } else {
+        numCols = null;
+      }
+      dispatch(addFacetLayout(FacetLayout({columns: numCols})));
+      // dispatch(addGroupFacet(GroupFacet({facet: {name: "facet", data: "cars_source_5", groupby: [field]}}), groupId)); // remove hardcoded data name
+      dispatch(addFacet({name: "facet",data: "5", groupby: field} as Facet, groupId));
```
ktbacher: How to lookup lyra ID for dataset associated with dropped field?

src/js/actions/facetLayoutActions.ts
```diff
@@ -0,0 +1,14 @@
+import {createStandardAction} from 'typesafe-actions';
+import {FacetLayoutRecord} from '../store/factory/FacetLayout';
+import {assignId} from '../util/counter';
+import {State} from '../store';
+import {Dispatch} from 'redux';
+
+export function addFacetLayout (payload: FacetLayoutRecord) {
```
jonathanzong, line 7: re: this file, facetLayoutReducer, and facetLayout.ts, i think they're related enough to core layout functionality that i would consider just adding the facet actions / reducer cases / record definitions to the existing layout files so that it's easier to find everything.

src/js/store/factory/FacetLayout.ts
```diff
@@ -0,0 +1,41 @@
+import {Map, Record, RecordOf} from 'immutable';
+
+/**
+ * Layouts align multiple groups
+ */
+export interface FacetLayout {
```
jonathanzong, line 6: is there enough overlap between FacetLayout and Layout that we can think about e.g.:
- a common interface that both of them "inherit" from via the typescript `&` (intersection type) operator
- a "parent type" defined as the `|` (union type) of the two?

i'm not suggesting either particular choice is more right here but consider whether or not that might simplify things in some places (it also might not simplify things at all, in which case feel free to reject this idea)

src/js/ctrl/export.ts
```diff
@@ -216,12 +227,16 @@ exporter.mark = function(state: State, internal: boolean, id: number) {
     spec.from = {data: facet.name};
   } else if (spec.from) {
     let fromId;
-    if ((fromId = spec.from.data)) {
+    if ((fromId = spec.from.name)) {
```
jonathanzong: i'm not a fan of assignment inside of a boolean expression for code readability purposes. i know you're just following the pattern that was already here, but would you mind re-writing these if cases to be more explicit?
e.g.
```
if (spec.from.name) {
 ...
}
else if (spec.from.data) {
  ... 
}
else if (spec.from.mark) {
 ...
}
else if (spec.from.facet.data) {
 ...
}

File: src/js/components/pipelines/FacetDropzone.tsx [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+  public handleDragOver = (evt) => {
+    if (evt.preventDefault) {
+      evt.preventDefault();
+    }
+
+    return false;
+  };
+
+  public handleDrop = ()  => {
+    this.props.facetField(this.props.dragging.fieldDef.name, this.props.groupId);
```
jonathanzong: `this.props.dragging.dsId` is the dataset id

src/js/components/pipelines/FacetDropzone.tsx
jonathanzong: see comment in `handleDrop` below

src/js/ctrl/export.ts
```diff
@@ -49,9 +49,20 @@ export function exporter(internal: boolean = false): Spec {
...
+  }
   return spec;
 }
 
+exporter.layouts = function (state: State, internal: boolean) {
+  const facetLayouts = state.getIn(['vis', 'present', 'facetLayouts']);
+  const layout = clean(duplicate(facetLayouts), internal);
+  const id = Object.keys(layout)[Object.keys(layout).length -1];
+  return layout[id];
+}
```
jonathanzong: if i'm understanding this correctly, the logic is:
- if there are any facet layouts in the lyra store, set the last one created as the layout in the exported spec

a few questions to help my understanding:
- why do we discard the previous ones?
- what are the cases where there will be more than one facet layout?
- how is this different from how non-facet layouts are handled?

jonathanzong:
awesome work! i left some comments, questions, and suggestions. exiting to see this getting so close!

src/js/ctrl/export.ts
ktbacher: Really wasn't sure on how the logic of the export worked so this might be something we can walk through tomorrow

src/js/store/factory/FacetLayout.ts
ktbacher: Interesting, they are similar in concept but very different in implementation and how they are used in Lyra and the ultimate vega spec.  Perhaps worth discussing further

src/js/store/index.ts
```diff
@@ -52,6 +54,7 @@ const getDefaultState = Record<LyraState>({
       widgets: Map<string, WidgetRecord>(),
       signals: defaultSignalState,
       layouts: Map<string, LayoutRecord>(),
+      facetLayouts: Map<string, FacetLayoutRecord>(),
```
jonathanzong, line 57: suggestion per our conversation today: since there can only be one vega layout spec, change this from a Map to a single FacetLayoutRecord

ktbacher: Probably just not understanding typescript syntax/types here, but when I change the line to `facetLayouts: FacetLayoutRecord` I get an error:`'FacetLayoutRecord' only refers to a type, but is being used as a value here.`

jonathanzong: Here you're initializing the actual values of a record, not defining an interface. So what you're trying to do in this line is set the value of a property `facetLayouts` to a default FacetLayoutRecord. So it should be `facetLayouts: FacetLayoutRecord()` -- note the parentheses for initialization

ktbacher: hmm I'm getting the same error even with the parentheses

jonathanzong: i meant `FacetLayout()`, for the constructor and not the type

ktbacher: ohhh make sense.  realizing with the proposed change to facet layouts to make it compatible with the other layout mechanism, we could in theory support multiple facets and might want to keep it as a map after all?

PR #543, opened 2021-05-04
Esefah meng drop zones

NanaKwame:
Pushing this for feedback


src/js/components/history/HistoryList.tsx
```diff
@@ -0,0 +1,267 @@
...
+class BaseHistoryList extends React.Component<OwnProps & StateProps & DispatchProps> {
+
+  public handleExpand = (e, historyId: number) => {
+    e.target.parentElement.classList.toggle("expanded");
+  }
+  public render() {
+    const isExpanded = this.props.expandedLayers[1000]; // todo(ej): have a better id for this state. use history state
+    const groupClass = isExpanded ? 'expanded' : 'contracted';
+
+    return (
```
NanaKwame, line 97: @jonathanzong this is where I suspect the issue is

